PR #6, opened 2016-03-08
Add selector-no-redundant-parent, closes #3

### LINKED ISSUE #3: Rule idea: selector-no-redundant-parent
Basically this: https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#unnecessaryparentreference

kristerkari:
Most of the tests are from https://github.com/brigade/scss-lint/blob/master/spec/scss_lint/linter/unnecessary_parent_reference_spec.rb

**Checklist**
- [x] Add readme

### DIFF
--- a/src/rules/selector-no-redundant-parent/index.js
+++ b/src/rules/selector-no-redundant-parent/index.js
@@ -1,0 +1,27 @@
+import { utils } from "stylelint"
+
+export const ruleName = "selector-no-redundant-parent"
+
+export const messages = utils.ruleMessages(ruleName, {
+  rejected: "Unnecessary parent selector (&)",
+})
+
+export default function (actual) {
+  return function (root, result) {
+    const validOptions = utils.validateOptions(result, ruleName, { actual })
+    if (!validOptions) { return }
+
+    root.walkRules(rule => {
+      const unnecessaryAmpersandRegex = /^&(\s+)?>?(\s+)?(\w+|\s\.)/
+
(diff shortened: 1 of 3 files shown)


src/rules/selector-no-redundant-parent/index.js
```diff
@@ -0,0 +1,27 @@
...
+
+export default function (actual) {
+  return function (root, result) {
+    const validOptions = utils.validateOptions(result, ruleName, { actual })
+    if (!validOptions) { return }
+
+    root.walkRules(rule => {
+      const unnecessaryAmpersandRegex = /^&(\s+)?>?(\s+)?(\w+|\s\.)/
+
+      if (rule.selector.trim() === "&" || unnecessaryAmpersandRegex.test(rule.selector.trim())) {
```
davidtheclark: Maybe you'd consider it too much of an edge case, but what if there's an `&` in an attribute, e.g. `div[id="foo&bar"]`?

Also: have you looked into postcss-selector-parser? Maybe not necessary if your regexp is working great, but might turn out more accurate in the end if you start running into edge cases. Just an idea.

kristerkari: > Maybe you'd consider it too much of an edge case, but what if there's an & in an attribute, e.g. div[id="foo&bar"]?

It's easy to add a test for that :)

> Also: have you looked into postcss-selector-parser? Maybe not necessary if your regexp is working great, but might turn out more accurate in the end if you start running into edge cases. Just an idea.

I'll definitely have a look, let's see if it makes sense to use it.

PR #7, opened 2016-03-25
Add extend-require-placeholder, closes #1

### LINKED ISSUE #1: Rule idea: require a placeholder when using @extend
Sometimes you only use `@extend` to share code between two or more classes, but never actually use the class that your extending. In this case it would be a good idea to require `@extend` to only be used with placeholder selectors (e.g. `%my-class`).

For example, this code:

```scss
.icon {
  transition: background-color ease .2s;
  margin: 0 .5em;
}

.error-icon {
  @extend .icon;
  /* error specific styles... */
}

.info-icon {
  @extend .icon;
  /* info specific styles... */
}
```

...gets compiled to:

```scss
.icon, .error-icon, .info-icon {
  transition: background-color ease .2s;
  margin: 0 .5em;
}

.error-icon {
  /* error specific styles... */
}

.info-icon {
  /* info specific styles... */
}
```

You have `.icon` defined in your CSS even if you don't plan to use it.

However, if you use a placeholder, this:

```scss
%icon {
  transition: background-color ease .2s;
  margin: 0 .5em;
}

.error-icon {
  @extend %icon;
  /* error specific styles... */
}

.info-icon {
  @extend %icon;
  /* info specific styles... */
}
```

...gets compiled to:

```scss
.error-icon, .info-icon {
  transition: background-color ease .2s;
  margin: 0 .5em;
}

.error-icon {
  /* error specific styles... */
}

.info-icon {
  /* info specific styles... */
}
```

All examples taken from:
http://thesassway.com/intermediate/understanding-placeholder-selectors

kristerkari:
Many of the tests and descriptions are taken from:
https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#placeholderinextend

### DIFF
--- a/src/rules/extend-require-placeholder/index.js
+++ b/src/rules/extend-require-placeholder/index.js
@@ -1,0 +1,28 @@
+import { utils } from "stylelint"
+
+export const ruleName = "extend-require-placeholder"
+
+export const messages = utils.ruleMessages(ruleName, {
+  rejected: "Use a placeholder selector (e.g. %some-placeholder) with @extend",
+})
+
+export default function (actual) {
+  return function (root, result) {
+    const validOptions = utils.validateOptions(result, ruleName, { actual })
+    if (!validOptions) { return }
+
+    root.walkAtRules("extend", atrule => {
+      const isPlaceholder = (/^%/).test(atrule.params.trim())
+      const isInterpolation = (/^#{.+}/).test(atrule.params.trim())
+
+      if (!isPlaceholder && !isInterpolation) {
+        utils.report({
+          ruleName,
+          result,
+          node: atrule,
+          message: messages.rejected,
+        })
+      }
+    })
+  }
+}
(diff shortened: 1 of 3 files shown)


src/rules/extend-require-placeholder/index.js
```diff
@@ -0,0 +1,28 @@
...
+  rejected: "Use a placeholder selector (e.g. %some-placeholder) with @extend",
+})
+
+export default function (actual) {
+  return function (root, result) {
+    const validOptions = utils.validateOptions(result, ruleName, { actual })
+    if (!validOptions) { return }
+
+    root.walkAtRules("extend", atrule => {
+      const isPlaceholder = (/^%/).test(atrule.params.trim())
```
davidtheclark: This would probably be some degree faster if you just checked the first character instead of running a regexp (`atrule.params.trim()[0] === '%'`)

kristerkari: True!

PR #9, opened 2016-04-19
Add dollar-variable-no-missing-interpolation, closes #2

### LINKED ISSUE #2: Rule idea: Require interpolation when string variables are used as custom identiefiers
Some CSS features like animations allow the user to specify a custom identifier for animation name. 

So if you have `animation-name` as a string inside a Sass variable, you need to use interpolation to get rid of the quotes. There could be a rule that requires interpolation to be used in those cases.

So this SCSS:

```scss
$var: "my-anim";

.class {
  animation-name: $var;
}

@keyframes $var {
}
```

...gets compiled to:

```css
.class {
  animation-name: "my-anim";
}

@keyframes "my-anim" {}
```

Which does not work. You need to use interpolation to get rid of the quotes:

```scss
$var: "my-anim";

.class {
  animation-name: #{$var};
}

@keyframes #{$var} {
}
```

...and that gets compiled to:

```css
.class {
  animation-name: my-anim;
}

@keyframes my-anim {}
```

More info:
- `animation-name` - MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/animation-name
- custom CSS identifier - MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/custom-ident

kristerkari:
**TODO**
- [x] Add readme


jeddy3:
Like I said in https://github.com/kristerkari/stylelint-scss/pull/16, I'm not sure how closely you want to follow the [stylelint conventions](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/about-rules.md#about-rule-names). If you do want to follow them though, this rule should probably be a `no` rule maybe called `dollar-variable-no-missing-interpolation` or `dollar-variable-no-valid-interpolation`.

src/rules/variable-validate-interpolation/index.js
```diff
@@ -0,0 +1,127 @@
+import { utils } from "stylelint"
+import valueParser from "postcss-value-parser"
+
+export const ruleName = "variable-validate-interpolation"
+
+export const messages = utils.ruleMessages(ruleName, {
+  rejected: (n, v) => `You need to interpolate your variable ${v} when using it with ${n}`,
```
jeddy3: FYI, messages in stylelint [start with](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#determine-warning-messages) "Expected ... " or "Unexpected ..." e.g. `"Expected variable \"${variable}\" to be interpolated when using it with \"${node}\""`

jeddy3:
FYI, READMEs in stylelint [a follow set format](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#write-the-readme). 

The one from [`dollar-variable-pattern`](https://github.com/kristerkari/stylelint-scss/blob/master/src/rules/dollar-variable-pattern/README.md), is a good one to copy and paste :)

src/rules/variable-validate-interpolation/index.js
kristerkari: Thanks! I need to check that the other rules follow the same message format.

kristerkari:
> Like I said in #16, I'm not sure how closely you want to follow the stylelint conventions. If you do want to follow them though, this rule should probably be a no rule maybe called dollar-variable-no-missing-interpolation or dollar-variable-no-invalid-interpolation.

I renamed this rule in a00511e.

kristerkari:
I decided to change the name to `dollar-variable-no-missing-interpolation`, because it is describes the rule better.

Issue #8, opened 2016-03-31
Rule idea: dollar-variable-pattern


arnaudvalle:
Following my question on SO (http://stackoverflow.com/questions/36304921) I'm proposing to add a rule to enforce a naming convention for variable names in SCSS.

Something like https://github.com/sasstools/sass-lint/pull/257 would be great I reckon, where we could either specify a string for a preferred common format (camelcase, snakecase etc...) or a custom pattern.

For example I may want my variable names to always be camel case, in which case:

```
// this wouldn't be correct
$my-variable: #000;

// but this would be correct
$myVariable: #000;
```
I don't have experience in writing plugins for stylelint but I may give it a try based on the other rules in this repo and try to work something out in a PR if it's of any use to anybody else.

jeddy3:
> I don't have experience in writing plugins for stylelint but I may give it a try

Hi @arnaudvalle. I'm not a contributor here, but I help out on the `stylelint` linter itself. Perhaps I can be of some help here...

There's some stylelint documentation that you might find useful when getting started: [working on rules](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md) & [writing plugins](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/plugins.md).

There's also some documentation on [naming rules](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules.md#about-rule-names) and [how rules are built to work together](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#rules-work-together).

Finally, there are a handful of similar `*-pattern` rules in stylelint that might provide a good starting point this plugin:

* [`custom-property-pattern`](https://github.com/stylelint/stylelint/blob/master/src/rules/custom-property-pattern/) - probably the most similar to this proposed plugin.
* [`custom-media-pattern`](https://github.com/stylelint/stylelint/tree/master/src/rules/custom-media-pattern)
* [`selector-class-pattern`](https://github.com/stylelint/stylelint/tree/master/src/rules/selector-class-pattern)

...

You might want to follow the same convention and call this plugin `variable-pattern`?

Good luck :)

arnaudvalle:
@jeddy3 thanks for the links, they'll definitely be useful when I have a go at this!

kristerkari:
Thanks! This is something that should not be that hard to implement.

jeddy3:
[...]

PR #10, opened 2016-04-28
Add `dollar-variable-pattern` rule

dryoma:
Implements #8

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "url": "https://github.com/kristerkari/stylelint-scss/issues"
   },
   "dependencies": {
+    "lodash": "^4.11.1",
     "stylelint": "^4.5.0"
   },
   "devDependencies": {
--- a/src/rules/dollar-variable-pattern/index.js
+++ b/src/rules/dollar-variable-pattern/index.js
@@ -1,0 +1,35 @@
+import { isRegExp, isString } from "lodash"
+import { utils } from "stylelint"
+
+export const ruleName = "dollar-variable-pattern"
+
+export const messages = utils.ruleMessages(ruleName, {
+  expected: "Expected $ variable name to match specified pattern",
+})
+
+export default function (pattern) {
+  return (root, result) => {
+    const validOptions = utils.validateOptions(result, ruleName, {
+      actual: pattern,
+      possible: [ isRegExp, isString ],
+    })
+    if (!validOptions) { return }
+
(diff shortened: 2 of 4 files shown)


dryoma:
That's strange. The log from Travis CI says 
> Cannot find module 'lodash'

But running

```
npm install
npm run test
```
on my local Node.js v 4.4.3 works fine.

package.json
```diff
@@ -12,6 +12,7 @@
     "url": "https://github.com/kristerkari/stylelint-scss/issues"
   },
   "dependencies": {
+    "lodash": "^4.11.1",
```
kristerkari: It looks like this is not being used anywhere? Let's not add lodash dependency unless it is really needed.

src/rules/dollar-variable-pattern/index.js
```diff
@@ -0,0 +1,35 @@
...
+    })
+    if (!validOptions) { return }
+
+    const regexpPattern = (isString(pattern))
+      ? new RegExp(pattern)
+      : pattern
+
+    root.walkDecls(decl => {
+      const { prop } = decl
+      if (prop.slice(0, 1) !== "$") { return }
```
kristerkari: This could also be written as `if (prop[0]) !== "$") { return }`, but I don't think that it makes much difference :)

package.json
kristerkari: Oh sorry, it is actually used. Ignore this comment.

kristerkari:
Thanks, good job with this rule! I'll try to finish my own PR, so that 1.0 can be shipped.

src/rules/dollar-variable-pattern/index.js
dryoma: True. Just tested `slice` to make sure, it is about 20% slower, but indeed for 100000 operations it makes only about 20ms in modern browsers. Yet, not the best code, sorry, my bad

dryoma:
Thanks, glad I could help.

src/rules/dollar-variable-pattern/index.js
kristerkari: No worries, it is easy to change if we want the 20% perf boost :)

arnaudvalle:
@dryoma - thanks for implementing this! I didn't get any time to have a look at it since creating the initial issue but it looks like you took care of it, so cheers!

Issue #12, opened 2016-04-28
Rule idea: mixin-name-pattern, function-name-pattern


dryoma:
Or `mixin-pattern`, `function-pattern` to follow the existing stylelint [rule naming conventions](http://stylelint.io/user-guide/rules/).

stylelint-scss already has a rule for placeholder names, the one for variable names is in a PR now. The subject rules are the ones missing. SCSS-lint has those (though in a format of [one general naming rule for Sass stuff](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#nameformat)). I could take implementing them, seems not that hard.

The only question I have: should some general patterns like `lowercase-underscore`, etc. be imlpemented as predefined primary option values (like in SCSS-lint)? If yes, what predefined patterns would you recommend to use?

dryoma:
Another thing that came to mind is maybe these rules should have `sass-`/`scss-` prefix? So that they don't conflict with other rules written for other preprocessors, and also don't misguide users on what preprocessor/technology they rely.

kristerkari:
> Another thing that came to mind is maybe these rules should have sass-/scss- prefix?

Yes, that can be done before releasing v1.0. I'll add that to the release checklist.

kristerkari:
> The only question I have: should some general patterns like lowercase-underscore, etc. be imlpemented as predefined primary option values (like in SCSS-lint)? If yes, what predefined patterns would you recommend to use?

Does stylelint use default options in rules? I think that we should try to keep things consistent with it and not with SCSS-lint.

dryoma:
I thougth that too (stylelint doesn't have predefined patterns for its core rules), but decided to suggest just in case.

jeddy3:
> Does stylelint use default options in rules? 

As you both spotted, there are no default primary options for stylelint rules. It was a conscience decision to both keeps the rules unopinionated and to mitigate having to deliberate on what the defaults options should be :)

jeddy3:
> Another thing that came to mind is maybe these rules should have sass-/scss- prefix?

This is tricky as I suspect quite a few of these rules will be applicable to [precss](https://github.com/jonathantneal/precss) users (or some other future processor). Hence the original suggestion of `dollar-variable-pattern`, rather than `scss-variable-pattern`. There could also be a plugin for Less users called `at-variable-pattern`.

I dunno if there's a right answer here though.

kristerkari:
[...]

PR #15, opened 2016-04-28
Add rules: scss-function-pattern, scss-mixin-pattern

dryoma:
Implements #12

### DIFF
--- a/src/rules/scss-function-pattern/index.js
+++ b/src/rules/scss-function-pattern/index.js
@@ -1,0 +1,36 @@
+import { isRegExp, isString } from "lodash"
+import { utils } from "stylelint"
+
+export const ruleName = "scss-function-pattern"
+
+export const messages = utils.ruleMessages(ruleName, {
+  expected: "Expected SCSS function name to match specified pattern",
+})
+
+export default function (pattern) {
+  return (root, result) => {
+    const validOptions = utils.validateOptions(result, ruleName, {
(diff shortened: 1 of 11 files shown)


src/rules/scss-function-pattern/index.js
```diff
@@ -0,0 +1,36 @@
...
+      possible: [ isRegExp, isString ],
+    })
+    if (!validOptions) { return }
+
+    const regexpPattern = (isString(pattern))
+      ? new RegExp(pattern)
+      : pattern
+
+    root.walkAtRules(declaration => {
+      const decl = declaration
```
kristerkari: I think that the code would be a bit more readable if you could just change `declaration` to `decl` and do something like this:

```js
root.walkAtRules(decl => {
  if (decl.name !== "function") { return }

  const funcName = decl.params.replace(/(\s*?)\(.*?\)/g, "")
  if (regexpPattern.test(funcName)) { return }
```

dryoma: Good point, changed that part. In the code for these rules I relied on stylelint `-pattern` rules code a lot, so decided to leave that mostly unchanged.

kristerkari:
I merged another PR and now there is a conflict. Could you rebase your branch against master?

dryoma:
Sure, see #17

kristerkari:
Closing this, let's continue with #18

PR #17, opened 2016-05-03
Add rules: scss-function-pattern, scss-mixin-pattern (updated branch)

dryoma:
Implements #12

An updated (confilcts-free) version of #15


dryoma:
Updated the branch and rebased it to another one. Some commits already in the repo got into th PR, but git seems ok with that :) Will that go fine?

kristerkari:
> Some commits already in the repo got into th PR, but git seems ok with that

Hmm... I'm not sure. I think that merging this will cause the same commits to show twice in commit history. At least that's what has happened before. 

How did you rebase?

I usually do it this way:

1) Pull the latest changes (`git checkout master` before if you need to)
2) `git checkout mybranch` (you should use a custom branch instead of master)
3) `git rebase master`
4) Fix conflict(s)
5) `git add file-that-conflicts`
6) `git rebase --continue` (now there should be no conflicts left)
7) `git push origin mybranch -f` use force to overwrite history (the conflicting commits in your branch)

dryoma:
I did almost the same except for not including `-f` flag in 7) and pushing to my fork's master. Now I did exactly as you suggested, yet the commits not from my fork are still shown here...

dryoma:
Ok #18 should work.

kristerkari:
👍

kristerkari:
Closing this, let's continue with #18

PR #18, opened 2016-05-03
Add rules: scss-function-pattern, scss-mixin-pattern -- take 3

dryoma:
Implements #12 

Ended up creating a whole new branch from the point the differences started and `cherry-pick`-ing new commit.


kristerkari:
Looks good. What do you think, should we rename these rules based on the discussion here https://github.com/kristerkari/stylelint-scss/issues/12#issuecomment-215681497?

Should we ship this with 1.0 since the rules are already implemented?

dryoma:
I would answer positively to both questions :)

Jeddy gave a really good agrument about some other preprocessors usage. For SCSS it doesn't make it any worse, I think (in any case, there will be docs to all the rules, and a package name, which is stylelint-**scss**). And we already have the **dollar** in `dollar-variable-pattern`.

kristerkari:
I also renamed my rule in #9

PR #24, opened 2016-05-11
Prepare version 1.0.0

kristerkari:
Prepare version 1.0.0. 

I have released a "pre-release" if you want to test it out, just `npm i stylelint-scss@0.0.0-alpha.1`.

ping @davidtheclark @jeddy3 @dryoma Let me know if there is anything that should be fixed before this plugin goes live.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,10 +7,6 @@
 
 A collection of SCSS specific linting rules for [stylelint](https://github.com/stylelint/stylelint) (in a form of a plugin).
 
-(Work in progress - not released yet).
-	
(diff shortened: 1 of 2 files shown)


dryoma:
I could only come up with adding `"prebuild": "rimraf dist",` to `"scripts"` (yet I'm not it is used in installing from NPM, more likely it's for Travis CI and other CIs).

kristerkari:
Ah true, I need to do that.

davidtheclark:
You got the namespace in there, so looks good.

kristerkari:
Anything else, or should we go ahead and release?

PR #25, opened 2016-05-20
Add `stylelint-plugin` keyword to `package.json`

ntwb:
Now that stylelint recommends adding `stylelint-keyword` to NPM packages to aid in discoverability of stylelint plugins: http://stylelint.io/developer-guide/plugins/#sharing-plugins


dryoma:
Thanks @ntwb !

Issue #19, opened 2016-05-03
Rule idea: at-rule-import-path


evilebottnawi:
https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#importpath

kristerkari:
Thanks! Sounds good, I'll add this to the next milestone.

jeddy3:
Perhaps called `at-import-no-invalid-partial-path`?

`at-rules-*` rules [tend to apply to all at-rules](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules.md#at-rule), rather than a specific type.

dryoma:
@jeddy3 I'm actually thinking of making this two separate rules: `at-import-no-partial-leading-underscore` and `at-import-no-partial-extension`. The last one might take an array of extensions to forbid (`["scss", "less"]`).

jeddy3:
> I'm actually thinking of making this into two separate rules: at-import-no-partial-leading-underscore and at-import-no-partial-extension.

Breaking this into two rules sounds good to me! You're getting better at this than me :)

> The last one might take an array of extensions to forbid (["scss", "less"]).

`no-*` rules tend to only have `true` as the primary options. So, in this instance, you can either use a blacklist rule or `no-` rule with [secondary option](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#determine-options) *that makes the rule more permissive* e.g. respectively:

- `at-import-partial-extension-blacklist: []`
- `at-import-no-partial-extension: [ true, { ignoreExtensions: ["less"] } ]` - where `ignoreExtensions` is a *user-defined* list of extensions to ignore.

I'm going to be on holiday for the next couple of weeks, and so I figured I'd feedback on these upcoming rules before I go. Great job with `1.0.0` btw :)

dryoma:
> Breaking this into two rules sounds good to me

Thanks. I'm just working on a project where Sass partials and bootstrap files (and even CSS) of the same name can be in the same directory, so we sometimes have to specifically tell Sass what exactly we are importing.

> `no-*` rules tend to only have true as the primary options

Sure, didn't pay attention to the name while writing. Yeah, the `no-` variant looks better to me (easier to set up). But since Sass treats `@import`-s with `.css` [as simple CSS directives](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#import), we'll have to ignore `.css` extension by default (I wouldn't like to force a user to add it manually).

[...]

PR #27, opened 2016-05-23
at-import-path- rules

dryoma:
Implements #19 

Decided to make it a PR from a separate branch as I believe there could be suggestions about the rules' names and functionality. If it's ok then I'll merge it (or you guys can, I don't mind :) )

### DIFF
--- a/src/rules/at-import-no-partial-extension/README.md
+++ b/src/rules/at-import-no-partial-extension/README.md
@@ -1,0 +1,74 @@
+# at-import-no-partial-leading-underscore
+
+Disallow file extensions in partial names in `@import`.
+
+```scss
+@import "path/to/file.scss"
+/**                  ↑
+ *                   Disallow this */
+```
+
+The rule ignores [cases](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#import) when Sass considers an `@import` command just a plain CSS import:
+
+* If the file’s extension is `.css`.
+* If the filename begins with `http://` (or any other protocol).
+* If the filename is a `url()`.
+* If the `@import` has any media queries.
+
+The following patterns are considered warnings:
+
+```scss
+@import "foo.scss";
+```
+
+```scss
+@import "path/fff.less";
+```
+
+```scss
+@import "path\\fff.supa";
+```
+
+```scss
+`@import "df/fff", '1.scss';`
+```
+
+The following patterns are *not* considered warnings:
+
+```scss
+@import "path/fff";
+```
+
+```scss
+@import url("path/_file.css"); /* has url(), so doesn't count as a partial @import */
+```
+
+```scss
+@import "file.css"; /* Has ".css" extension, so doesn't count as a partial @import */
+```
+
+```scss
+/* Both are URIs, so don't count as partial @imports */
+@import "http://_file.scss";
+@import "//_file.scss";
+```
+
+```scss
+@import "file.scss" screen; /* Has a media query, so doesn't count as a partial @import */
+```
+
+## Optional Options
+
+### `ignoreExtensions: [ string or regexp ]`
+
+An array of extensions to ignore, elements *don't need the dot at the start*. If an element is a string, only extensions that match that string are ignored. If an element is a regular expression, then extensions are tested against it and ignored if the test is successful.
+
+For example, with `whitelist: [ "scss", /^my/ ] ]`, the following are no longer warnings:
+
+```scss
+@import "path/fff.scss";
+```
+
+```scss
+@import "path/fff.my01";
+```
--- a/src/rules/at-import-no-partial-extension/README.md
+++ b/src/rules/at-import-no-partial-extension/README.md
@@ -1,4 +1,4 @@
-# at-import-no-partial-leading-underscore
+# at-import-no-partial-extension
 
 Disallow file extensions in partial names in `@import`.
 
@@ -30,7 +30,7 @@ The following patterns are considered warnings:
 ```
 
 ```scss
-`@import "df/fff", '1.scss';`
+@import "df/fff", '1.scss';
 ```
 
 The following patterns are *not* considered warnings:
@@ -63,7 +63,7 @@ The following patterns are *not* considered warnings:
 
 An array of extensions to ignore, elements *don't need the dot at the start*. If an element is a string, only extensions that match that string are ignored. If an element is a regular expression, then extensions are tested against it and ignored if the test is successful.
 
-For example, with `whitelist: [ "scss", /^my/ ] ]`, the following are no longer warnings:
+For example, with `ignoreExtensions: [ "scss", /^my/ ] ]`, the following are no longer warnings:
 
 ```scss
 @import "path/fff.scss";
--- a/src/rules/at-import-no-partial-extension/__tests__/index.js
+++ b/src/rules/at-import-no-partial-extension/__tests__/index.js
@@ -1,0 +1,179 @@
+import testRule from "stylelint-test-rule-tape"
+import rule, { ruleName, messages } from ".."
+
+// Testing main value
+testRule(rule, {
+  ruleName,
+  config: [true],
+  syntax: "scss",
+
+  accept: [ {
+    code: `
+      @import "fff";
+    `,
+    description: "Single file, no extension, double quotes.",
+  }, {
+    code: `
+      @import 'fff';
+    `,
+    description: "Single file, no extension, single quotes.",
+  }, {
+    code: `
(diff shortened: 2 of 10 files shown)


src/rules/at-import-no-partial-extension/README.md
```diff
@@ -0,0 +1,74 @@
+# at-import-no-partial-leading-underscore
```
kristerkari: Wrong title :)

src/rules/at-import-no-partial-extension/README.md
```diff
@@ -0,0 +1,74 @@
...
+```scss
+@import "path/fff.less";
+```
+
+```scss
+@import "path\\fff.supa";
+```
+
+```scss
+`@import "df/fff", '1.scss';`
```
kristerkari: Extra back-ticks.

src/rules/at-import-no-partial-extension/README.md
```diff
@@ -0,0 +1,74 @@
...
+@import "file.scss" screen; /* Has a media query, so doesn't count as a partial @import */
+```
+
+## Optional Options
+
+### `ignoreExtensions: [ string or regexp ]`
+
+An array of extensions to ignore, elements *don't need the dot at the start*. If an element is a string, only extensions that match that string are ignored. If an element is a regular expression, then extensions are tested against it and ignored if the test is successful.
+
+For example, with `whitelist: [ "scss", /^my/ ] ]`, the following are no longer warnings:
```
kristerkari: This should be `ignoreExtensions` instead of `whitelist`.

src/rules/at-import-no-partial-extension/__tests__/index.js
```diff
@@ -0,0 +1,179 @@
...
+  }, {
+    code: `
+      @import "fff.scss", 'fils.less';
+    `,
+    description: "Single file, extension string-ignored.",
+  }, {
+    code: `
+      @import "fff", "fff.scss";
+    `,
+    description: "Multiple files, on of the extensions string-ignored.",
```
kristerkari: Should this be `"Multiple files, on of the imports regex-ignored."` ?

src/rules/at-import-no-partial-extension/__tests__/index.js
```diff
@@ -0,0 +1,179 @@
...
+  }, {
+    code: `
+      @import "fff.scss";
+    `,
+    description: "Single file, extension string-ignored.",
+  }, {
+    code: `
+      @import "fff", "fff.scss";
+    `,
+    description: "Multiple files, on of the extensions string-ignored.",
```
kristerkari: `on` => `one`

src/rules/at-import-no-partial-extension/__tests__/index.js
```diff
@@ -0,0 +1,179 @@
...
+  accept: [ {
+    code: `
+      @import "fff";
+    `,
+    description: "Single file, no extension.",
+  }, {
+    code: `
+      @import "fff.scss", 'fils.less';
+    `,
+    description: "Single file, extension string-ignored.",
```
kristerkari: e.g. `"Multiple files, one of the extensions regex-ignored and one string-ignored."`

src/rules/at-import-no-partial-extension/index.js
```diff
@@ -0,0 +1,65 @@
...
+        // Accepting array of either strings or regular expressions
+        ignoreExtensions: str => isRegExp(str) || isString(str),
+      },
+      optional: true,
+    })
+    if (!validOptions) { return }
+
+    function checkPathForUnderscore(path, decl) {
+      // Stripping trailing quotes and whitespaces, if any
+      const pathStripped = path.replace(/^\s*?("|')\s*?/, "").replace(/\s*?("|')\s*?$/, "")
```
kristerkari: I think that there should be some tests that include whitespace that these regexes should match. The tests are passing if I leave only the part that replaces quotes:

```js
const pathStripped = path.replace(/("|')$/, "")
```

src/rules/at-import-no-partial-extension/index.js
```diff
@@ -0,0 +1,65 @@
...
+      // Skipping importing CSS: url(), ".css", URI with a protocol, media
+      if (pathStripped.slice(0, 4) === "url(" ||
+        pathStripped.slice(-4) === ".css" ||
+        pathStripped.search("//") !== -1 ||
+        pathStripped.search(/(?:\s|[,)"'])\w+$/) !== -1
+      ) {
+        return
+      }
+      // If the extension is not empty
+      if (!extension) { return }
```
kristerkari: You should probably have this check happen before checking for CSS importing:

```
no extension return late x 1,429,822 ops/sec ±1.40% (76 runs sampled)
no extension return early x 10,920,300 ops/sec ±1.03% (89 runs sampled)
Fastest is no extension return early
```

src/rules/at-import-no-partial-leading-underscore/README.md
```diff
@@ -0,0 +1,59 @@
...
+```scss
+@import "path/_fff";
+```
+
+```scss
+@import "path\\_fff"; /* Windows delimiters */
+```
+
+```scss
+`@import "df/fff", '_1.scss';`
```
kristerkari: Extra back-ticks.

src/rules/at-import-no-partial-leading-underscore/index.js
```diff
@@ -0,0 +1,44 @@
...
+      // Skipping importing CSS: url(), ".css", URI with a protocol, media
+      if (pathStripped.slice(0, 4) === "url(" ||
+        pathStripped.slice(-4) === ".css" ||
+        pathStripped.search("//") !== -1 ||
+        pathStripped.search(/(?:\s|[,)"'])\w+$/) !== -1
+      ) {
+        return
+      }
+      // Searching a _ at the start of filename
+      if (pathStripped.search(/(?:^|\/|\\)_[^/]+$/) === -1) { return }
```
kristerkari: Same here, there seems to be a speed boost if you do this before CSS import checks:

```
no leading _ search late x 1,170,845 ops/sec ±3.89% (77 runs sampled)
no leading _ search early x 13,380,512 ops/sec ±2.25% (79 runs sampled)
Fastest is no leading _ search early
```

src/rules/at-import-no-partial-leading-underscore/index.js
```diff
@@ -0,0 +1,44 @@
...
+})
+
+export default function (actual) {
+  return (root, result) => {
+    const validOptions = utils.validateOptions(result, ruleName, { actual })
+    if (!validOptions) { return }
+
+    function checkPathForUnderscore(path, decl) {
+      // Stripping trailing quotes and whitespaces, if any
+      const pathStripped = path.replace(/^\s*?("|')\s*?/, "").replace(/\s*?("|')\s*?$/, "")
```
kristerkari: Same as in the other rule, this needs tests.

src/rules/at-import-no-partial-extension/__tests__/index.js
dryoma: Thanks! Fixed that and other typos/misleading titles.

src/rules/at-import-no-partial-extension/index.js
dryoma: Sure, added the tests.

src/rules/at-import-no-partial-extension/index.js
dryoma: Also done.

By the way, what tool did you use to get those results? I though about adopting stylelint's `benchmark` script but left if for later.

kristerkari: I ran https://benchmarkjs.com/ using a copy/paste of the code. Using `benchmark-rule.js` stylelint is a good idea.

kristerkari:
👍 Looks good, unless there is something related to naming of the rules or other stylelint conventions.

dryoma:
@jeddy3 seems to be ok [with the naming](https://github.com/kristerkari/stylelint-scss/issues/19#issuecomment-220550712) and he said he won't be online for a couple of weeks. So i think I'll merge this for now and if we find something odd later we'll just fix that.

BRW thanks for such a detailed looking at the code. So much better to have more than one pair of eyes =)

jeddy3:
> @jeddy3 seems to be ok with the naming and he said he won't be online for a couple of weeks. 

Looks great to me! I'm properly back online in a week's time and so I'll answer any new questions you have then.  Keep up the good work btw, this plugin (pack) is excellent! :)

Issue #26, opened 2016-05-20
Rule Idea: selector-placeholder-pattern


mehcode:
https://github.com/stylelint/stylelint/pull/1338

kristerkari:
I guess we can create this since we already have other `-pattern` rules implemented.

dryoma:
I made it working in a draft version, yet I have some minor doubts before pushing.

1. Placeholders could be used both in extends (`.foo { @extend %placeholder; }`) and in selectors, where they are defined (`%placeholder { color: #000; }`). Do we need to process all these cases, or maybe having an `includeDefinitions` optional param would be better? The arguments for the param would be a) parsing selectors would take longer, and b) if allplaceholders in `@extend` are forced to match a pattern then corresponding placeholder definitions will have to comply either.

1. Do we need options like [`resolveNestedSelectors`](http://stylelint.io/user-guide/rules/selector-class-pattern/#resolvenestedselectors-true--false-default-false)? I think it is quite obvious that selectors must be resolved to allow for code like this:
    ```scss
    %foo {
      &-mod { ... }
    }
    ```
    Yet stylelint's `selector-class-pattern` does have it.

kristerkari:
> Placeholders could be used both in extends (.foo { @extend %placeholder; }) and in selectors, where they are defined (%placeholder { color: #000; }). Do we need to process all these cases, or maybe having an includeDefinitions optional param would be better?

I think that we should only validate the placeholder when it is defined. Isn't that what we do in the other rules too?

> Do we need options like resolveNestedSelectors? I think it is quite obvious that selectors must be resolved to allow for code like this:

That could be a good option to have, but probably not that critical for the first version of the rule.

dryoma:
> I think that we should only validate the placeholder when it is defined. Isn't that what we do in the other rules too?

It makes perfect sense. Dunno why I was focusing `@extend` here at all :)

PR #31, opened 2016-05-25
New rule: percent-placeholder-pattern. Some utility functions

dryoma:
Implements #26

### DIFF
--- a/src/rules/percent-placeholder-pattern/index.js
+++ b/src/rules/percent-placeholder-pattern/index.js
@@ -1,0 +1,70 @@
+import _ from "lodash"
+import resolveNestedSelector from "postcss-resolve-nested-selector"
+import { utils } from "stylelint"
+import {
+  hasInterpolatingAmpersand,
+  isStandardRule,
+  isStandardSelector,
+  parseSelector,
+} from "../../utils"
+
+export const ruleName = "percent-placeholder-pattern"
+
+export const messages = utils.ruleMessages(ruleName, {
+  expected: placeholder => `Expected %-placeholder "%${placeholder}" to match specified pattern`,
+})
+
+export default function (pattern) {
+  return (root, result) => {
+    const validOptions = utils.validateOptions(result, ruleName, {
+      actual: pattern,
+      possible: [ _.isRegExp, _.isString ],
+    })
+    if (!validOptions) { return }
+
+    const placeholderPattern = (_.isString(pattern))
+      ? new RegExp(pattern)
+      : pattern
+
+    // Checking placeholder definitions (looking among regular rules)
+    root.walkRules(rule => {
+      const { selector } = rule
+
+      // If it's a custom prop or a less mixin
+      if (!isStandardRule(rule)) { return }
+      // If the selector has interpolation
+      if (!isStandardSelector(selector)) { return }
+      
+      // Nested selectors are processed in steps, as nesting levels are resolved.
+      // Here we skip processing intermediate parts of selectors (to process only fully resolved selectors)
+      // if (rule.nodes.some(node => node.type === "rule" || node.type === "atrule")) { return }
+
+      // Only resolve selectors that have an interpolating "&"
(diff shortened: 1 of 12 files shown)


kristerkari:
I have a couple of review comments, but as it is late already, I'll return with those tomorrow.

kristerkari:
I was away for the weekend, so unfortunately I did not have time to add my comments yet.

src/rules/percent-placeholder-pattern/index.js
```diff
@@ -0,0 +1,70 @@
...
+    const placeholderPattern = (_.isString(pattern))
+      ? new RegExp(pattern)
+      : pattern
+
+    // Checking placeholder definitions (looking among regular rules)
+    root.walkRules(rule => {
+      const { selector } = rule
+
+      // If it's a custom prop or a less mixin
+      if (!isStandardRule(rule)) { return }
```
kristerkari: There should be a test for this check.

src/rules/percent-placeholder-pattern/index.js
```diff
@@ -0,0 +1,70 @@
...
+      // Here we skip processing intermediate parts of selectors (to process only fully resolved selectors)
+      // if (rule.nodes.some(node => node.type === "rule" || node.type === "atrule")) { return }
+
+      // Only resolve selectors that have an interpolating "&"
+      if (hasInterpolatingAmpersand(selector)) {
+        resolveNestedSelector(selector, rule).forEach(selector => {
+          parseSelector(selector, result, rule, s => checkSelector(s, rule))
+        })
+      } else {
+        parseSelector(selector, result, rule, s => checkSelector(s, rule))
```
kristerkari: This call is the same as above. It could be extracted into a function:

```js
function parse(selector) {
  parseSelector(selector, result, rule, s => checkSelector(s, rule)
}
resolveNestedSelector(selector, rule).forEach(parse)
```

src/rules/percent-placeholder-pattern/index.js
```diff
@@ -0,0 +1,70 @@
+import _ from "lodash"
```
kristerkari: Should we just import the functions that we need from lodash, like we do in other rules?

```js
import { isRegExp, isString } from "lodash"
```

kristerkari:
It would be good to have some tests for the utility functions too.

src/utils/hasInterpolatingAmpersand.js
```diff
@@ -0,0 +1,23 @@
...
+ * Check whether a selector has an interpolating ampersand
+ * An "interpolating ampersand" is an "&" used to interpolate within another
+ * simple selector (e.g. `&-modifier`), rather than an "&" that stands
+ * on its own as a simple selector (e.g. `& .child`)
+ *
+ * @param {string} selector
+ * @return {boolean} If `true`, the selector has an interpolating ampersand
+ */
+export default function (selector) {
+  for (let i = 0, l = selector.length; i < l; i++) {
```
kristerkari: Small thing, but there is really no need to save `selector.length` into a variable. There is no perf boost from that in modern JS engines.

kristerkari:
Nice test suite by the way. 👍

src/rules/percent-placeholder-pattern/index.js
dryoma: Sure. The used variant is a rudiment from when some util functions were inside this file.

src/rules/percent-placeholder-pattern/index.js
dryoma: The code below would rule out non-placeholder rules and declarations anyway, and this (and the following) checks are just early returns for not parsing the selector. Anyway, added a couple of tests for LESS.

dryoma:
Thanks for the review, I added the code changes. As for utils tests - I simply copied them from stylelint repo until they expose them for plugins (and *if* they do so). So the tests for them are basically in stylelint repo :)

PR #34, opened 2016-05-30
selector-no-redundant-nesting-selector: fix false positives, fixes #32

### LINKED ISSUE #32: selector-no-redundant-nesting-selector false positive when followed by a character other than ':'
'selector-no-redundant-nesting-selector' rule falsely assumes the selector is redundant in this common BEM use case:

```
.block {
  &__element {
    ...
  }
  &--modifier {
   ...
  }
}
```

More generally nesting selector is not redundant if immediately followed by a non white space character as it results in a new not nested rule.

kristerkari:
Fix false positives when `&` is used to combine classname from multiple parts or when BEM syntax is used.

### DIFF
--- a/src/rules/selector-no-redundant-nesting-selector/__tests__/index.js
+++ b/src/rules/selector-no-redundant-nesting-selector/__tests__/index.js
@@ -45,13 +45,6 @@ testRule(rule, {
   }, {
     code: `
       p {
-        & + .foo {}
-      }
-    `,
-    description: "when an ampersand precedes a sibling operator",
-  }, {
-    code: `
-      p {
         & + &:hover {}
       }
     `,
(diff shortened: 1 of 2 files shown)


dryoma:
By the way, congratulations with the first bug filed and fixed! :)

src/rules/selector-no-redundant-nesting-selector/__tests__/index.js
```diff
@@ -137,6 +160,53 @@ testRule(rule, {
...
+        & ~ .foo {}
+      }
+    `,
+    line: 3,
+    message: messages.rejected,
+    description: "when an ampersand precedes a sibling operator",
+  }, {
+    code: `
       p {
         &  .class {}
```
dryoma: `&  #class {}`, `&  %class {}`, `&  &-class {}` — these will not return errors.

kristerkari: ah, true! Need to add tests for those too.

dryoma: I pulled a fix for that: #40 If nothing is left out, you can merge it and publish a patch (as per a suggestion in #38

kristerkari: Alright, I'll do a patch release now.

PR #35, opened 2016-05-30
Fix severity not working because ruleName is not prefixed, fixes #33

### LINKED ISSUE #33: All the plugin rules are showing up severity: "ignore"
image

I'm having trouble where all the rules from this plugin are severity warning. I'm not sure if it's a stylelint system problem because the error on `border: none;` is from a plugin also.

I've also tried this in my config file

```json
    "scss/at-extend-no-missing-placeholder": [true, { "severity": "error" }],
```

And I still get a ignore

kristerkari:
A quick fix to have stylelint correctly report `error` or `warning` instead of `ignore` for stylelint-scss rules. 

It seems that the plugin name has to match rule's name for everything to work. We have only been namespacing rule names when `createPlugin` gets called, which resulted in bug #33.

Would be nice to know if there is a cleaner way of fixing this than calling `namespace` function in each rule. /CC @davidtheclark @jeddy3 @ntwb @dryoma

### DIFF
--- a/src/rules/at-extend-no-missing-placeholder/index.js
+++ b/src/rules/at-extend-no-missing-placeholder/index.js
@@ -1,6 +1,7 @@
 import { utils } from "stylelint"
+import { namespace } from "../../utils"
 
-export const ruleName = "at-extend-no-missing-placeholder"
+export const ruleName = namespace("at-extend-no-missing-placeholder")
 
 export const messages = utils.ruleMessages(ruleName, {
   rejected: "Expected a placeholder selector (e.g. %placeholder) to be used in @extend",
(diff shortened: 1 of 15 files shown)


jonrohan:
Oh cool! That wasn't an obvious fix. Thanks for getting to it. 👍

kristerkari:
I took me a while to figure out what is going on :)

davidtheclark:
Yeah, each rule has to be namespaced like that. The way that you've done it looks smart to me :)

src/rules/at-extend-no-missing-placeholder/index.js
```diff
@@ -1,6 +1,7 @@
 import { utils } from "stylelint"
+import { namespace } from "../../utils"
 
-export const ruleName = "at-extend-no-missing-placeholder"
+export const ruleName = namespace("at-extend-no-missing-placeholder")
```
dryoma: I think we should update the *Contributing* section of the docs to require namespacing names of new rules, and doing it exactly this way (no hardcoded namespace, no other utils, etc.).

kristerkari: Sure thing, I'll update my PR.

Issue #28, opened 2016-05-24
Rule idea: at-import-disallow-in-partials


dryoma:
Importing partials from other partials can lead to difficulties following what is included where, and lead to mistakes of duplicate imports (met those just about a month ago). Grouping all `@import` commands in Sass bootstrap files (those that get compiled into .css) on the other hand may improve clarity as all the includes are in one place, and the order is clear and strict.

Hugo Giraudel [elaborates on the approach](https://sass-guidelin.es/#architecture) in a greater detail.

kristerkari:
So would this rule check against `_` prefix in filename + `@import` inside the file?

dryoma:
Indeed.

kristerkari:
Implemented in #42

PR #42, opened 2016-06-04
New rule: partial-no-import

dryoma:
Implements #28 

Decided to change the name since it is partials that seem to be the *thing* (tha THIIIING!) here, not imports.

Was also happy to be able to peep a test example (not only scss code, but filenames have to be checked here) in stylelint repo :)

### DIFF
--- a/src/rules/partial-no-import/index.js
+++ b/src/rules/partial-no-import/index.js
@@ -1,0 +1,55 @@
+import { utils } from "stylelint"
+import { namespace } from "../../utils"
+import nodeJsPath from "path"
+
+export const ruleName = namespace("partial-no-import")
+
+export const messages = utils.ruleMessages(ruleName, {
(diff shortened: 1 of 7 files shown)


dryoma:
Ok, merging this. If there are any quirks, please add comments to commit code.

src/rules/partial-no-import/index.js
```diff
@@ -0,0 +1,55 @@
...
+        ruleName,
+      })
+    }
+
+    // Usual CSS file
+    if (extName === ".css") { return }
+    // Not a partial
+    if (fileName[0] !== "_") { return }
+
+    root.walkAtRules("import", decl => {
```
jeddy3: Should `decl` be `atRule` here?

jeddy3:
Besides that one variable name query, this looks great to me!

> it is partials that seem to be the thing (tha THIIIING!) here

:)

Issue #41, opened 2016-06-03
add new excludeExtensions blacklist option for `at-import-no-partial-extension`


rob3c:
Rule `at-import-no-partial-extension` currently isn't very useful when namespacing files like `foo.bar.scss`, `foo.baz.scss`, etc. That's because stylelint-scss sees an import like

`@import "foo.bar";` // <-- bar should not be flagged as an extension here

as having extension `bar` instead of recognizing that it isn't using a common `.scss`, `.less`, etc. known extension and ignoring it.

The stylelist-scss API can handle this, in principle, since you can pass blacklist `RegExp`s like `/^(?!scss)/` in the `ignoreExtensions` array when using the API in javascript. However, the JSON format used by `.stylelistrc` doesn't support `RegExp`s natively and only allows strings.

Rather than try to have stylelist-scss figure out what's a string vs a regex, I think the most direct way to address this would be to add a new config option `excludeExtensions` that would act as an extension blacklist similar to how `ignoreExtensions` acts as a whitelist.

dryoma:
To be honest, I did think about this use case, but decided to leave it for later. So no excuse to chill now :)

dryoma:
What if we split this one into two rules: `at-import-partial-extension-blacklist` and `at-import-partial-extension-whitelist` instead of two options to a "binary" rule? Will this be too much fragmentation?

@jeddy3 @kristerkari what do you think?

jeddy3:
> What if we split this one into two rules: at-import-partial-extension-blacklist and at-import-partial-extension-whitelist instead of two options to a "binary" rule? Will this be too much fragmentation?

I think that's the most idiomatic way to go about this.

kristerkari:
Done in https://github.com/kristerkari/stylelint-scss/pull/46

rob3c:
Thanks for adding this! It looks like you've also addressed issues with URIs and url() that I hadn't originally considered but make perfect sense.

PR #46, opened 2016-06-12
New rules: at-import-partial-extension-blacklist, at-import-partial-extension-whitelist

dryoma:
Implements #41 

Two separate rules for extensions blacklist and whitelist, the old one gets deprecated (and to be removed in 2.0.0).

There is a problem though. Deprecation warning makes all the tests for `at-import-no-partial-extension` fail. Is there a more graceful way of giving such warnings?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Added: `media-feature-value-dollar-variable` rule.
 - Added: `at-import-partial-extension-blacklist` rule.
 - Added: `at-import-partial-extension-whitelist` rule.
+- Deprecated: `at-import-no-partial-extension` rule.
 
 # 1.1.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Added: `partial-no-import` rule.
 - Added: `media-feature-value-dollar-variable` rule.
 - Added: `at-import-partial-extension-blacklist` rule.
+- Added: `at-import-partial-extension-whitelist` rule.
 
 # 1.1.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Added: `partial-no-import` rule.
 - Added: `media-feature-value-dollar-variable` rule.
+- Added: `at-import-partial-extension-blacklist` rule.
 
 # 1.1.1
 
--- a/src/rules/at-import-no-partial-extension/index.js
+++ b/src/rules/at-import-no-partial-extension/index.js
@@ -43,7 +43,7 @@ export default function (on, options) {
       if (options && options.ignoreExtensions) {
         // Return if...
         if (options.ignoreExtensions.some(ignoredExt => {
-          // the extension matches on of the ignored strings or Regexps
+          // the extension matches one of the ignored strings or Regexps
           return isString(ignoredExt) && ignoredExt === extension ||
             isRegExp(ignoredExt) && extension.search(ignoredExt) !== -1
         })) { return }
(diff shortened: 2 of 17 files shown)


src/rules/at-import-no-partial-extension/index.js
```diff
@@ -23,6 +23,13 @@ export default function (on, options) {
     })
     if (!validOptions) { return }
 
+    result.warn((
```
dryoma: This one bothers me. The idea was to give users a warning that this rule is deprecated. But the downside is that all the rule's tests are failing (each is giving +1 warning).

jeddy3: We'd normally just commented out all the tests and then we'd [add a test or two for the deprecation warning](https://github.com/stylelint/stylelint/blob/master/src/rules/font-family-name-quotes/__tests__/index.js#L269). That felt like the best balance between moving the project forward and using our time wisely.

src/rules/at-import-partial-extension-blacklist/README.md
```diff
@@ -0,0 +1,70 @@
+# at-import-partial-extension-blacklist
+
+Specify blacklist of disallowed file extensions for partial names in `@import` commands.
+
+```scss
+@import "path/to/file.scss"
+/**                  ↑
+ *                   Blacklist of these */
```
jeddy3: Minor niggle, but the the convention is keep things as far left as possible.

If we don't expect the dot to be included then I suggest pointing the arrow at the "s" or "scss", and then aligning the bottom of arrow with the last "e" in "these".

jeddy3:
Looking good!

src/rules/at-import-no-partial-extension/index.js
dryoma: Yea, that hits the spot, thanks!

jeddy3: Nice. I just did the [same thing](https://github.com/stylelint/stylelint/pull/1430) over at stylelint.

jeddy3:
This is looking structurally sound to me. I'll leave @kristerkari to comb through it though for you.

kristerkari:
I'll have a look at this now :)

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Head
 
 - Added: `partial-no-import` rule.
+- Added: `at-import-partial-extension-blacklist` rule.
```
kristerkari: Changelog also needs to mention the deprecated rule.

src/rules/at-import-partial-extension-whitelist/README.md
```diff
@@ -0,0 +1,78 @@
...
+The rule ignores [cases](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#import) when Sass considers an `@import` command just a plain CSS import:
+
+* If the file’s extension is `.css`.
+* If the filename begins with `http://` (or any other protocol).
+* If the filename is a `url()`.
+* If the `@import` has any media queries.
+
+## Options
+
+`array`: `"["array", "of", "extensions"]`
```
kristerkari: Extra `"` before the array.

src/rules/at-import-partial-extension-blacklist/README.md
```diff
@@ -0,0 +1,70 @@
...
+The rule ignores [cases](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#import) when Sass considers an `@import` command just a plain CSS import:
+
+* If the file’s extension is `.css`.
+* If the filename begins with `http://` (or any other protocol).
+* If the filename is a `url()`.
+* If the `@import` has any media queries.
+
+## Options
+
+`array`: `"["array", "of", "extensions"]`
```
kristerkari: Extra `"` before the array.

kristerkari:
Thanks for implementing these rules! Looks good. Do we want to ship these with next minor release?

dryoma:
Thanks @kristerkari 

I think we can safely publish this as no breaking changes have been introduced (I added two new rules and the old one is almost untouched except for the deprecation warning). And `at-import-no-partial-extension` is to be removed in the next major release. So I'll leave this for you to merge if you agree (all the issues you pointed at are fixed).

Issue #29, opened 2016-05-24
Rule idea: at-media-no-non-variable / at-media-no-non-dollar-variable


dryoma:
... or something like this. The point is to force all values inside `@media` queries be variables. The `-dollar-` variant is to underline that the rule forces `$var`s, not custom properties or variables from some other syntaxes.

Examples:
```scss
// Good:
@media (max-width: $width-mid) { ... }

// Bad:
@media (max-width: 1024px) { ... }
```

kristerkari:
Why not, could be useful. 👍

dryoma:
I have some cases for this rule I'd like to discuss. Suppose we have it as `media-query-value-dollar-variable: "always"/"never"`, which by itself looks ok. Then:

```scss
@media (max-width: 1024px) { ... } // Warning for "always"
@media (max-width: $var) { ... } // Warning for "never"

@media (max-width: 1024px + $var) { ... } // False positive for "never"
@media (max-width: $var + 1024px) { ... } // Not even sure how to treat this

@media (max-width: someFunc()) { ... } // Same here
```

1) We might need some additional rules for `calculation` and `function` as a media query value. 2) Or just process those in this rule when using some additional options like `ignoreCalculation`. The first option seems decent at first, but not sure at all about a rule just prohibiting a math operation in a value. The second one would keep it all compact, but might result in too many options.

What woul be your thoughts guys?

kristerkari:
My opinion is that if the rule is called `media-query-value-dollar-variable`, then it should only care about dollar variables, so in my opinion:
- For `always` only allow `$var` and `#{$var}` (because it would be difficult to allow any kind of calculation with all kinds of edge cases).
- For `never` don't allow any `$var` or calculation that contains `$var`, but allow calculation like `1000px + 24px` because it does not contain a variable.

I just noticed when writing this that that is exactly how you have implemented that rule, so 👍

kristerkari:
Implemented in https://github.com/kristerkari/stylelint-scss/pull/47

PR #47, opened 2016-06-13
New rule: media-query-value-dollar-variable

dryoma:
Implements #29 

Since this rule is about forcing/disallowing `$`-variables, the logic ended up like this:

* `always`: a value must be a single variable, possibly inside interpolation, nothing else is allowed.
* `never`: a value must not have variables at all (function calls are allowed, but not with variables as their arguments).


kristerkari:
Should the rule be `media-feature-*` instead `media-query-*` like the rules are in stylelint? 

https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules.md#media-feature

kristerkari:
Hmm.. there is also 1 rule in media query category:
https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules.md#media-query

src/rules/media-query-value-dollar-variable/index.js
```diff
@@ -0,0 +1,62 @@
...
+export default function (expectation, options) {
+  return function (root, result) {
+    const validOptions = utils.validateOptions(result, ruleName, {
+      actual: expectation,
+      possible: [ "always", "never" ],
+    }, {
+      actual: options,
+      optional: true,
+      possible: {
+        ignore: [ "calculations", "funcitons" ],
```
kristerkari: These options are not implemented? There is a typo too `"funcitons"`.

src/rules/media-query-value-dollar-variable/index.js
```diff
@@ -0,0 +1,62 @@
...
+      },
+    })
+    if (!validOptions) { return }
+
+    // In `(max-witdh: 10px )` find `: 10px )`
+    const valueRegex = /:(?:\s*?)(\S.+?)(:?\s*?)\)/
+    const valueRegexGlobal = new RegExp(valueRegex.source, "g")
```
kristerkari: Is there any reason to use the regex constructor here? You could just put the `g` option to the regex literal.

src/rules/media-query-value-dollar-variable/index.js
```diff
@@ -0,0 +1,62 @@
...
+      if (!found || !found.length) { return }
+      found.forEach(function (found) {
+        // ... parse it to "10px"
+        const valueParsed = found.match(valueRegex)[1]
+
+        // A value should be a single variable without any appendices (math operations)
+        // or it should be a single variable inside Sass interpolation
+        if (expectation === "always" && !(valueParsed[0] === "$" &&
+          valueParsed.search(/(\s|[+-/*])/) === -1 ||
+          valueParsed.search(/^#\{\s*?\$[A-Za-z_0-9]+\s*?\}$/) !== -1
```
kristerkari: The code is probably a bit more readable if you save these regexes into variables.

kristerkari: Here you are matching a variable as `\$[A-Za-z_0-9]+`, but for `never` option it is just `\$`. Maybe there should be a unified way to match a variable.

src/rules/media-query-value-dollar-variable/index.js
kristerkari: Ah, you are actually using both of them. Ignore this then.

src/rules/media-query-value-dollar-variable/index.js
```diff
@@ -0,0 +1,62 @@
...
+    }, {
+      actual: options,
+      optional: true,
+      possible: {
+        ignore: [ "calculations", "funcitons" ],
+      },
+    })
+    if (!validOptions) { return }
+
+    // In `(max-witdh: 10px )` find `: 10px )`
```
kristerkari: Typo in this comment `witdh` :)

src/rules/media-query-value-dollar-variable/index.js
```diff
@@ -0,0 +1,62 @@
...
+            message: messages.expected,
+          })
+        } else if (expectation === "never" && valueParsed.search(/\$/) !== -1) {
+          // "Never" means no variables at all (functions allowed)
+          utils.report({
+            ruleName,
+            result,
+            node: atRule,
+            word: valueParsed,
+            message: messages.rejected,
```
kristerkari: You could change both of these calls to `utils.report` to a single function and pass it the message as a parameter.

jeddy3:
> Should the rule be media-feature-* instead media-query-* like the rules are in stylelint?

Yep, I think it should be

> Hmm.. there is also 1 rule in media query category:

Hmmm, indeed. That's a legacy thing where at some point I was convinced that you could wrap groups of media features in parens (so `media-query-parentheses-space-inside` would target the media feature parens *and the surrounding ones*) e.g:

```css
@media ((color) and (min-width: 100px), (max-width: 20px)) {}
```

Now I'm no longer sure if that's right.

Anyhow, defo `media-feature-value-dollar-variable` for this rule.

kristerkari:
> Anyhow, defo media-feature-value-dollar-variable for this rule.

Thanks for clarifying 👍

src/rules/media-query-value-dollar-variable/index.js
jeddy3: Idiomatically, [we call](jeddy3/linter-stylelint) that function `complain`.

src/rules/media-query-value-dollar-variable/index.js
dryoma: Yeah, I wasn't sure whether to use options up till the last moment (even after pushing). Now since we 100% agree on how this rule should work I can purge them.

src/rules/media-query-value-dollar-variable/index.js
dryoma: The point is that with "always" the value must be one solid variable name, so we have to check for that. And for "never" just locating a `$` sign means there is a variable somewhere there, and we can safely throw a warning since we don't care what else is there in the value.

By the way, I cleaned the criteria a bit.

dryoma:
Renamed the rule. Hope the terminology doesn't change %)

Issue #20, opened 2016-05-03
Rule-idea:space-around-operator


evilebottnawi:
https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#spacearoundoperator

kristerkari:
Thanks! I guess this could be useful for someone, so I'll add this to the next milestone.

jeddy3:
Perhaps called `calculation-no-unspaced-operator` as "calculation" seems to be a [*thing*](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/about-rules.md#about-rule-names) in SCSS) so it can be used in tandem with stylelint's [core rule](https://github.com/stylelint/stylelint/tree/master/src/rules/function-calc-no-unspaced-operator)

dryoma:
@jeddy sure, no problem.

As about the rule. I am goint to make it exactly a `-no-` rule, meaning prohibiting spaces around operators seems to be no-go. At least because `$var--lol` is a valid variable name in Sass, and if we prohibit spaces we will force users to write `$var - 10px` as `$var-10px`, which would lead to a totally incorrect result.

I'd also save ourselves an effort distinguishing `calc` from all the other functions and would rather check operators inside `calc()` here as well.

jeddy3:
> I am goint to make it exactly a -no- rule

I agree. Well reasoned.

evilebottnawi:
i think best options: "always" and "never".

dryoma:
@evilebottnawi with `"never"` users won't be able to write:

* `$var - 10px`; they will be forced to write it as `$var-10px`, which Sass interprets as a variable name (like in `$var-10px: #faaac0;`)
* `font-family: sans- + "serif";`; they will have to do it as `font-family: sans - + "serif";` which is compiled into `font-family: sans-+"serif";`

evilebottnawi:
@dryoma oops, i forget ;) right rule must have `no`

dryoma:
Some more things I'd like to bring up:

1. Make `operator` rather than `calculation` the thing here, as the rule would work with conditions and logical operations (`@if ($i == 10 and $j != 1)`) as well.
[...]

PR #55, opened 2016-07-10
Rules to control spaces and newlines around Sass operators

dryoma:
Implements #20 

Along with borrowing some existing stylelint utils I had to add two complex tools: one for pasring media queries and another for parsing strings for Sass operators.

The media parser, I suppose, is better off as a standalone tool. I'm just not sure whether it should be prepared as some kind of a PostCSS-stuff (like [postcss-value-parser](https://github.com/TrySound/postcss-value-parser)), and whether there are any requirements for a standalone tool to be called PostCSS-<something> at all. So, any thought are more than welcome.

As for operations parser: the more I added to it the more I grew to believe we'll eventually need a real full fledged Sass parser. At least the one for operations. Cause all those helpers do help in simple cases, but in complex ones like now they'll just grow too ugly.

The rules' intended mechanics were already described in the [issue](https://github.com/kristerkari/stylelint-scss/issues/20), and to sum it up:
* `operator-no-unspaced` only allows to **force** whitespace around operators, and on both sides. It is because `$var-10` and `$var - 10` are different things in Sass;
* `operator-no-newline-` can prohibit newlines but don't require them. That is to allow users to place multiple operands on the same line. Just like ESLint's [operator-linebreak](http://eslint.org/docs/rules/operator-linebreak).

-------

**TODO:**
- [x] Resolve interpolation in CSS block comments
- [x] Remove media query parser util, replace with `postcss-media-query-parser`.


jeddy3:
Wow. Awesome job!

> The media parser, I suppose, is better off as a standalone tool. 

Yes please. This has been on our bucket list for a while! At least one core rule proposal that will use it springs to mind.

> I'm just not sure whether it should be prepared as some kind of a PostCSS-stuff

I'm not sure either.

kristerkari:
Thanks! I'll try to review this soonish.

dryoma:
@jeddy3 got it. Do you maybe have requests for some other features (other than `each` and `walk` methods) for the media parser?

jeddy3:
> Do you have requests for some other features (other than each and walk methods) for the media parser?

Nope, those two would be super!

dryoma:
@jeddy3 done: https://github.com/dryoma/postcss-media-query-parser (will later remove it from the code here).

jeddy3:
Awesome, thanks!

src/rules/operator-no-unspaced/README.md
```diff
@@ -0,0 +1,96 @@
...
+a { width: "10*1"; }} // Inside a string, ignored
+```
+
+```scss
+// Linebreak will do as a whitespace; indentation before "-" and trailing spaces after "1" are left to the corresponding stylelint rules
+a {
+  width: 1  
+    - a;
+}
+```
```
kristerkari: Maybe there should also be some examples like these to show what is accepted:

```scss
a { width: 10 * 1; }
```

```scss
a { width: 10 + 1; }
```

src/rules/operator-no-unspaced/README.md
```diff
@@ -0,0 +1,96 @@
...
+```scss
+a { width: (10) /#{1}; } // Has interpolation on one of the sides, so not an operation
+```
+
+```scss
+a { width: $var-1; } // "$var-1" is a variable name
+```
+
+```scss
+a { width: "10*1"; }} // Inside a string, ignored
```
kristerkari: Extra `}`

src/rules/operator-no-unspaced/README.md
dryoma: Sure, I was so busy dealing with edge cases that usual cases ended up completely discriminated against %)

dryoma:
Added: internal media parsing tool, resolving interpolation in CSS comments; exapmles to readme. And fixed some typos.

Any more comments on the code here? Or shall I merge this?

BTW @jeddy3 do you think stylelint might also need a [comment parser](https://github.com/kristerkari/stylelint-scss/blob/13a73323619ac186f184d1704c1b7ad19c0a411b/src/utils/findCommentsInRaws.js) to be able to check comments that are dropped by postcss at the moment? i.e. should it also declare independence.

src/rules/operator-no-unspaced/index.js
```diff
@@ -0,0 +1,237 @@
...
+
+    // Check a value (`10px` in `width: 10px;`)
+    if (item.value !== undefined) {
+      results.push({
+        source: item.value,
+        operators: findOperators({
+          string: item.value,
+          globalIndex: declarationValueIndex(item),
+          // For Sass variable values some special rules apply
+          isAfterColon: item.prop[0] === "$" ? true : false,
```
kristerkari: This could be just `isAfterColon: item.prop[0] === "$"`, no need for ternary.

src/utils/sassValueParser/index.js
```diff
@@ -0,0 +1,788 @@
...
+ * @return {Array} array of { symbol, globalIndex, startIndex, endIndex }
+ *    for each operator found within a string
+ */
+export default function findOperators({
+  string,
+  globalIndex,
+  isAfterColon,
+  callback,
+}) {
+  // console.log("===============================================")
```
kristerkari: Maybe remove this? :)

src/utils/sassValueParser/index.js
```diff
@@ -0,0 +1,788 @@
...
+  const isValueWithUnitBefore_ = isValueWithUnitBefore(before)
+  const isNumberAfter_ = isNumberAfter(after)
+  const isNumberBefore_ = isNumberBefore(before)
+  const isInterpolationAfter_ = isInterpolationAfter(after)
+  const isParensAfter_ = isParensAfter(after)
+  const isParensBefore_ = isParensBefore(before)
+  // The early check above helps prevent deep recursion here
+  const isPrecedingOperator_ = isPrecedingOperator(string, index)
+  
+  // console.log(string, index, string[index])
```
kristerkari: This could also be removed.

kristerkari:
I added a couple of small comments, but other than that I think that this is good to merge.

There is quite a lot of code with all the needed utils/parsers, but there are also a ton of tests to verify that things work correctly.

src/rules/operator-no-unspaced/index.js
```diff
@@ -0,0 +1,237 @@
...
+        }),
+      })
+      match = interpolationRegex.exec(string)
+    }
+    return results
+  }
+
+  root.walk(item => {
+    // console.log('-----------------------')
+    // console.log(item)
```
kristerkari: Maybe remove these console logs too.

jeddy3:
> BTW @jeddy3 do you think stylelint might also need a comment parser, to be able to check comments that are dropped by postcss at the moment?

We [decided](https://github.com/stylelint/stylelint/pull/1683#issuecomment-234721247) just to document this limitation and wait for `postcss@6`. We already have three `comment-*` rules in core. Adding support for selector and value lists to these rules would be a breaking change. So, we figured we'd keep the limitation consistent across the old and new rules, and wait to fix them all with `postcss@6`.

dryoma:
Ok, merging this. And waiting for a heckish flood of bug reports %)

> We decided just to document this limitation and wait for postcss@6
Well, I thought it'd be not that productive to wait for the 6th version (and IF they fix the comments in it), since we have a specific rule for forcing/prohibiting // comments to be inline, and the tool was not that difficult to implement.

Issue #52, opened 2016-06-29
Rule idea: variable-name-space-after


bjankord:
This rule idea comes from scss-lint's [Space After Variable Name rule](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#spaceaftervariablename)

From the scss-lint docs:

_Variables should be formatted with no space between the name and the colon._

**Bad: space before colon**
```scss
$my-var : 0;
```

**Good**
```scss
$my-var: 0;
```

dryoma:
Thanks, definitely a must.

In the meantime, stylelint's [`declaration-colon-space-before`](http://stylelint.io/user-guide/rules/declaration-colon-space-before/) and other `declaration-colon-` rules could help a bit, with variables declared inside rulesets.

jeddy3:
> In the meantime, stylelint's `declaration-colon-space-before` and other `declaration-colon-*` rules could help a bit, with variables declared inside rulesets.

It's interesting that `declaration-colon-space-before` only ignores unnested dollar variables at the moment. You could argue that it is should ignore them all, and that it is an oversight that it doesn't already. Either way, I doubt we'll change the behaviour until you guys have:

- `dollar-variable-colon-space-after`
- `dollar-variable-colon-space-before`

Which would check the spaces around dollar variable colons, whether nested or not. As, as you said, it's useful at the moment :)

dryoma:
Published with 1.3.0.

Issue #51, opened 2016-06-29
Rule idea: variable-colon-space-after


bjankord:
This rule idea comes from scss-lint's [Space After Variable Colon rule](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#spaceaftervariablecolon)

From the scss-lint docs:

_Variables should be formatted with a single space separating the colon from the variable's value._

**Bad: no space after colon**
```scss
$my-color:#fff;
```

**Bad: more than one space after colon**
```scss
$my-color:  #fff;
```

**Good**
```scss
$my-color: #fff;
```

dryoma:
Published with 1.3.0.

PR #56, opened 2016-07-13
Rules for whitespaces around colon in $-variable declarations

dryoma:
Implements #51, #52.

* `dollar-variable-colon-space-after`
* `dollar-variable-colon-space-before`
* `dollar-variable-colon-newline-after`


kristerkari:
Awesome work again! I've been quite busy this week, but I'll try to have a look at this and the other PR tomorrow.

src/rules/dollar-variable-colon-newline-after/index.js
```diff
@@ -0,0 +1,62 @@
...
+  namespace,
+  whitespaceChecker,
+} from "../../utils"
+import { utils } from "stylelint"
+
+export const ruleName = namespace("dollar-variable-colon-newline-after")
+
+export const messages = utils.ruleMessages(ruleName, {
+  expectedAfter: () => "Expected newline after \":\"",
+  rejectedAfter: () => "Unexpected whitespace after \":\"",
```
kristerkari: This message does not seem to be used.

src/rules/dollar-variable-colon-space-after/README.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+### `"never"`
+
+There *must never* be whitespace after the colon.
+
+The following patterns are considered warnings:
+
+```scss
+$var:10px;
```
kristerkari: Why would this be a warning? There is no whitespace after the colon here.

src/rules/dollar-variable-colon-newline-after/index.js
```diff
@@ -0,0 +1,62 @@
...
+      actual: expectation,
+      possible: [
+        "always",
+        "always-multi-line",
+      ],
+    })
+    if (!validOptions) { return }
+
+    root.walkDecls(decl => {
+      if (!decl.prop || decl.prop[0] !== "$") { return }
```
kristerkari: Maybe we should create a `isDollarVariable` function like this?

```js
function isDollarVariable(prop) {
  return typeof prop === "string" && prop[0] === "$";
}
```

src/rules/dollar-variable-colon-newline-after/index.js
```diff
@@ -0,0 +1,62 @@
...
+      // Get the raw prop, and only the prop
+      const endOfPropIndex = declarationValueIndex(decl) + decl.raw("between").length - 1
+
+      // The extra characters tacked onto the end ensure that there is
+      // a character to check after the colon (like in `$var:\n10px` - there
+      // would otherwise be nothing after `$var:` on that line)
+      const propPlusColon = decl.toString().slice(0, endOfPropIndex) + "xxx"
```
kristerkari: hmm.. this feels a bit dirty, any other options to do it?

src/rules/dollar-variable-colon-newline-after/index.js
bjankord: +1

src/rules/dollar-variable-colon-newline-after/index.js
dryoma: Indeed it seems so. I was initially relying on `declaration-colon-newline-after` rule from stylelint, so I thought it could be. Will remove the message and correct the rule's description.

src/rules/dollar-variable-colon-newline-after/index.js
dryoma: I'm not really sure. Such check won't be sufficient for cases like [this](https://github.com/kristerkari/stylelint-scss/blob/master/src/rules/media-feature-value-dollar-variable/index.js#L25) where the check for whether it's a $-variable has to be more thorough. Won't the proposed util name be misguiding?

src/rules/dollar-variable-colon-newline-after/index.js
dryoma: In fact there is. whitespaceChecker util was imported from stylelint, so was this hack initially. I wasn't sure if trying to fix this won't break the code. Well, for our rules it doesn't :) so appliying it.

src/rules/dollar-variable-colon-newline-after/index.js
kristerkari: Yeah, you are right. Let's not do it if it does not work in all cases.

kristerkari:
Looks good, should we merge?

src/rules/dollar-variable-colon-space-before/index.js
```diff
@@ -0,0 +1,34 @@
+import {
+  namespace,
+  whitespaceChecker,
+} from "../../utils"
+import { utils } from "stylelint"
+import { variableColonSpaceChecker } from "../dollar-variable-colon-space-after"
```
kristerkari: hmm.. Should `variableColonSpaceChecker` be refactored out of the rule? This creates tight coupling between the rules.

dryoma: Oops, sorry, merged before this comment appeared on the page.

As for the question - the rules' logic is ought to be very similar (actually, almost the same except for where the space is checked and the messages). It is quite a normal practice in stylelint (like [here](https://github.com/stylelint/stylelint/blob/master/src/rules/declaration-colon-space-before/index.js#L15)), and our rules use it too (like the neighbouring sass operators rules).

Issue #44, opened 2016-06-06
Rule idea: declaration-nested-properties


s10wen:
From https://github.com/stylelint/stylelint/issues/1386#issuecomment-223788453 @jeddy3 suggested I post this here.

kristerkari:
Good idea 👍

dryoma:
Getting down to this at last.

And the question. Do we really want the "always" option here? That would mean this rule will have to enforce nesting properties that... which ones exactly?

1) Definitely not those with a `-` in a name, 'cause `text-align` and stuff. Or no?

```scss
text: {
  align: center;
  decoration: underline;
  indent: 2em;
}
```

2) Or should we probably create a list of rules that are not allowed in non-nested form? E.g. `[ /^margin-/, /^flex-/, /^background-/, etc.]`, and if the parser meets one of them it yields something like `Expected "margin-left" to be in a form of a nested rule`.

3) Or just `"declaration-no-nested-properties`?

s10wen:
@dryoma interesting, regarding `2)` are you thinking of setting defaults, or have the user set them? Also wondering what would be the best approach, blacklist / whitelist, or give the user the option to set them, then in the docs show a suggestion.

kristerkari:
3) would be the easiest to implement, but yeah 2) could also work as set of whitelisted properties

dryoma:
If we go with whitelist/blacklist, than no defaults are possible by the nature of stylelint. I was hoping that someone could point at the list of properties that are usually/preferably use nesting in Sass, as Sass itself (according to [Sassmeister](http://www.sassmeister.com/)) does'n seem to give two moos about what is nested:

```scss
a {
  may: {
    the: {
      force: be with you;
    }
  }
}
```
```css
a {
  may-the-force: be with you;
}
```

kristerkari:
I guess if Sass does not care about what you nest, maybe we should not either -- and just skip option 2.

So we could do 3), or 1) that would have `"always" ` option to enforce nesting for any property that has `-` in the name.

dryoma:
If we go with 1), then is this:

```scss
a {
  display: block;
  margin: {
    left: 10px;
  }
}
```
acceptable? I.e. nesting just one prop. Maybe this rule must be triggered only if two or more props with the same prefix (`margin-`) are met?

kristerkari:
or maybe require everything to be nested by default and add  an option `except: first`?

dryoma:
So, with `"always"`: 
[...]

PR #63, opened 2016-07-27
Rules for nested properties

dryoma:
Implements #44 

* [`declaration-nested-properties`](https://github.com/kristerkari/stylelint-scss/blob/nested/src/rules/declaration-nested-properties/README.md)
* [`declaration-nested-properties-no-divided-groups`](https://github.com/kristerkari/stylelint-scss/blob/nested/src/rules/declaration-nested-properties-no-divided-groups/README.md).

And also a [parseNestedPropRoot](https://github.com/kristerkari/stylelint-scss/blob/nested/src/utils/parseNestedPropRoot.js) tool. At the moment PostCSS considers this:
```scss
font: 10px/1 Arial {
  weight: 500;
}
```
a **rule**, so `font: 10px/1 Arial` goes as a `selector` property in a corresponding object. The parseNestedPropRoot tool would help to find out whether a selector is actually such declaration and check it with our linting rules as well. Now I suppose such things cause false positives and false negatives (will check the rules later).


src/rules/declaration-nested-properties/README.md
```diff
@@ -0,0 +1,144 @@
...
+a background:red {
+  color: blue;
+}
+
+*/
+```
+
+## Optional options
+
+### `except: ["only-of-namespace"]`
```
kristerkari: I think that the option would be easier to understand if it was called `"only-one-of-namespace"` instead of `"only-of-namespace"`. What do you think?

kristerkari:
This looks good! Rebasing is needed, and is there something else that needs to be done?

src/rules/declaration-nested-properties/README.md
dryoma: There is [:only-of-type](http://reference.sitepoint.com/css/pseudoclass-onlyoftype) pseudo class, so I thought that naming would be ok (afiak this is ok grammatically either). But no personal feelings here. We need a third party to settle this %)

dryoma:
Thanks :)

But there is still one thing bothering me (that's why I wasn't merging for that long): the fact that
```scss
font: 10px/1 Arial {
  weight: 500;
}
```
is parsed as
```js
{
  type: 'rule',
  selector: 'font: 10px/1 Arial',
  ...
}
```
is 1) just bizzare, 2) requires hacks in these (dedicated) rules, 3) would possible require hacks in other rules that might need to distinguish Rule from Declaration.

I tried pushing a PR to postcss-scss, but in the end the result is not much better (I'd say worse) than what we have now. So what would you say if we use a fork of postcss-scss as a dependency?

kristerkari:
> I tried pushing a PR to postcss-scss, but in the end the result is even worse than what we have now. So how about we use a fork of postcss-scss as a dependency?

Hmm.. let's think about that. Forks are kind of annoying to maintain (since it usually diverges from the main repo), but we might have to do that if there is no better way.

dryoma:
Considering a change in ther parser for `margin: { $var: 1; left: $var; }` (so that it is not a Rule but a Declaration) is in question,

**The hack**
Pros:
+ We have full control over it and usually don't need to keep up with some other code.

Cons:
- Have to be applied in every instance where nested props could come up.
- Additional level of parsing stuff that is already parsed by PostCSS.
- (kinda questionable) If PostCSS does change, the hack will have to be updated.

**A fork for postcss-scss**
Pros:
+ Will make nested props parsed by postcss with no need in further complex checks (only basic ones, e.g. `if (item.isNested) {`.
+ Less processing work for cases like [this](https://github.com/kristerkari/stylelint-scss/blob/8ba304c2d1e0749a4625a19e8966dc3dc40f7ef1/src/rules/percent-placeholder-pattern/index.js#L31)

Cons:
- Will have to be updated every time the original project gets updated.
- On the other hand, in cases like [this](https://github.com/kristerkari/stylelint-scss/blob/366977a6367d0cfb5ce949c18d564689d60a029d/src/rules/dollar-variable-no-missing-interpolation/index.js#L82) and [here](https://github.com/kristerkari/stylelint-scss/blob/366977a6367d0cfb5ce949c18d564689d60a029d/src/rules/dollar-variable-no-missing-interpolation/index.js#L66) additional checks might get necessary.

---

Additional moments:

```scss
$var: margin;

a {
  #{$var}: 10px { // Here a variable _has_ to have interpolation, otherwise it's an error
    origin: left;
  }
}
```

dryoma:
Ok, merging this for now.

PR #72, opened 2016-08-29
Fix failing partial-no-import test

kristerkari:
This fixes a test that is currently failing on master branch. 

ping @dryoma

### DIFF
--- a/src/rules/partial-no-import/index.js
+++ b/src/rules/partial-no-import/index.js
@@ -50,6 +50,8 @@ export default function (on) {
     if (extName === ".css") { return }
     // Not a partial
     if (fileName[0] !== "_") { return }
+    // Filename starting with "_" and including "\"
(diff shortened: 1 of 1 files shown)


dryoma:
Hmm, that's strange. Everything is ok on my side. Maybe it has something to do with whether it's windows/linux and the slashes used?

kristerkari:
> Hmm, that's strange. Everything is ok on my side. Maybe it has something to do with whether it's windows/linux and the slashes used (I use Windows)?

That might be, I'm using osx and it is also failing on my machine.

dryoma:
The problem was in hard-coded path slashes I used to test dirs starting with `_`. I'm afraid `if (/\\/.test(fileName)) { return }` won't help here.

I pushed a fix in #73

kristerkari:
ok, thanks! Let's close this then.

PR #75, opened 2016-09-07
Testing stuff

dryoma:
* Replace `babel-tape-runner` with the original `tape`.
    `babel-tape-runner` uses `babel-polyfill`. So if a code has constructs like `Object.assign` which a JS interpreter (e.g. Node v0.12) doesn't, tests will run flawlessly, but actual code runs will throw.
    
    `tape` can "require" external modules befor running tests now, so it's safer to use `tabe -r babel-register`. And I also love ditching big complex stuff for small vanilla stuff :)
    
    The original issue: https://github.com/wavded/babel-tape-runner/issues/15 Thanks @olegskl for investigating.
* Update such Node v0.12 non-compatible code.
* Add [`tap-spec`](https://github.com/scottcorgan/tap-spec) for prettifying the output.
* Add helper scripts for easier test running
    E.g. `npm run test-rule operator-no-unspaced` will run
    
    ```
    node_modules/.bin/tape -r babel-register src/rules/partial-no-import/__tests__/indes.js | node_modules/.bin/tap-spec
    ```
    
    (so tired of typing this all the time)
* update the link to the project's NPM module in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # stylelint-scss
 
-[![NPM version](https://img.shields.io/npm/v/stylelint-scss.svg)](https://www.npmjs.org/package/stylelint-scss)
+[![NPM version](https://img.shields.io/npm/v/stylelint-scss.svg)](https://www.npmjs.com/package/stylelint-scss)
 [![Build Status](https://travis-ci.org/kristerkari/stylelint-scss.svg?branch=master)](https://travis-ci.org/kristerkari/stylelint-scss)
 [![Build status](https://ci.appveyor.com/api/projects/status/xa12kju6qmvmqs1n/branch/master?svg=true)](https://ci.appveyor.com/project/kristerkari/stylelint-scss/branch/master)
 [![Coverage Status](https://coveralls.io/repos/github/kristerkari/stylelint-scss/badge.svg?branch=master)](https://coveralls.io/github/kristerkari/stylelint-scss?branch=master)
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "rimraf": "^2.5.2",
     "semver": "^5.1.0",
     "stylelint-test-rule-tape": "^0.2.0",
+    "tap-spec": "^4.1.1",
     "tape": "^4.6.0"
   },
   "engines": {
@@ -78,6 +79,6 @@
     "prepublish": "npm run build",
     "pretest": "npm run lint",
     "release": "npmpub",
-    "test": "nyc tape -r babel-register \"src/**/__tests__/*.js\""
+    "test": "nyc tape -r babel-register \"src/**/__tests__/*.js\" | tap-spec"
   }
 }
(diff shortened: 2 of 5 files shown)


kristerkari:
Nice! I tried running `npm run test-rule operator-no-unspaced`, but I'm getting:
```
> node ./scripts/test-rule.js "operator-no-unspaced"

env: node: No such file or directory
env: node: No such file or directory
```

dryoma:
This might have something to do with how Node is being installed in the OS. Would any of [these solutions](http://stackoverflow.com/questions/20886217/browserify-error-usr-bin-env-node-no-such-file-or-directory) work for you? Seems that calling `node <script>.js` is [not that uncommon](https://github.com/eslint/eslint/blob/master/package.json#L11) in the community, I doubt all those people use Windows =)

kristerkari:
Nope, that's not it. The error is coming from `{ env: { FORCE_COLOR: "always" } }`.

dryoma:
Sorry, didn't have an opportunity to look into this.

Good thing is I fixed that, at least works in Ubuntu 16.

And bad news: Node.js 0.12 tests started to fail. At first it was stylelint's `table` dependency that adopded babel compiling for Node >=4. [Using its fixed version](https://github.com/kristerkari/stylelint-scss/pull/75/commits/b2fcd06807af67f2bd593becb3ac19addb085027) helped, but now this module strikes a pose:
```
/home/travis/build/kristerkari/stylelint-scss/node_modules/stylelint/node_modules/cosmiconfig/index.js:3
const path = require('path');
^^^^^
SyntaxError: Use of const in strict mode.
```
I suppose stylelint guys don't want anything below v4 anymore. So, not that I want to, but do we have an option beside doing the same since 1) this project havily uses stylelint, 2) it's by it nature won't be used without stylelint?

kristerkari:
Yeah I think that it's a good time to drop 0.12 support already. At least it seems that everyone else is doing that too.

dryoma:
I suppose we don't have to wait until the next major release? because stylelint is already 0.12-incompatible

kristerkari:
Hmm.. good point. If we don't do it in a major release, then we could break projects and builds for some people. It is possible that someone uses an older version of Stylelint and the current version of stylelint-scss if they need to still use 0.12

dryoma:
But won't our dependency of `"stylelint": "^7.0.3"` break their code? And not just that, but stylelint's dependencies. E.g. `table` did that breaking change in a minor release.

kristerkari:
You are right, stylelint has already dropped 0.12 support in that version, so we should do it too.

kristerkari:
@dryoma I opened a PR to drop 0.12 support https://github.com/kristerkari/stylelint-scss/pull/77

kristerkari:
@dryoma Now that you are changing some things related to tests, how about also adding test coverage? I think that it would be useful for us to know if there are untested things or unused code.

stylelint's coverage PR:
https://github.com/stylelint/stylelint/pull/1279

kristerkari:
I also went and added coverage reporting, so you need to rebase against master.

PR #78, opened 2016-10-20
Add coverage reporting using nyc & coveralls

kristerkari:


src/rules/operator-no-unspaced/index.js
```diff
@@ -159,7 +159,7 @@ export function calculationOperatorSpaceChecker({ root, result, checker }) {
 
       // Media queries
       if (item.name === "media" || item.name === "import") {
-        mediaQueryParser(item.params).walk(node => {
+        mediaQueryParser(item.params).walk("", (node) => {
```
kristerkari: I added this change because there is a bug in `mediaQueryParser`'s `walk` method that causes it to throw an error when running unit tests and coverage reporting to fail.

dryoma: Can't publish the patch, some problems with NPM's servers. Will do as soon as they get fixed.

kristerkari: @dryoma I tried installing `postcss-media-query-parser` 0.2.2 to remove this workaround, but it seems that you forgot to rebuild before releasing it to npm, and only the `src` folder has the correct changes.

dryoma: Shoot, I really did. Forgot to add a `prepublish` script. Thanks for pointing that out!

dryoma:
Nice work adding this, btw. 

> Coverage: 94%
So... we rock? :)

PR #79, opened 2016-10-31
Remove media query parser's walk method workaround

kristerkari:


dryoma:
Thanks, and sorry for the trouble.

BTW, since the check passes, I think this one could just have been commited into master ;)

Issue #53, opened 2016-07-02
Rule idea: dollar-variable-empty-line-before


jeddy3:
- Name: `dollar-variable-empty-line-before`
- Primary option: `"always"|"never"`
- Secondary options: `{ "except": ["first-nested", "after-comment", "after-dollar-variable"] }`
- Messages: 
  - Rejected: "Unexpected empty line before dollar-variable"
  - Expected: "Expected empty line before dollar-variable"
- Section: "`$`-variable"

Given:

``` js
[ "always", { "except": ["first-nested", "after-comment", "after-declaration"] }]
```

The following is valid:

``` css
a {
  $var1: value; /* "always" except first-nested */
  $var2: value; /* "always" except after-declaration */

  /* comment */
  $var3: value; /* "always" except after-comment */

  @extends "x";

  $var4: value; /* primary option is "always" */

   & b {
     prop: value;
   }

   $var5: value;  /* primary option is "always" */
}
```

Given:

``` js
"never"
```

The following is valid:

``` css
a {
  $var: value; /* primary option is "never" */
  $var: value; /* primary option is "never" */

  /* comment */
  $var: value; /* primary option is "never" */

  @extends "x";
  $var: value; /* primary option is "never" */

   & b {
     prop: value;
   }
   $var: value; /* primary option is "never" */
}
```

``` console
1:1 Unexpected empty line before dollar variable (scss/dollar-variable-empty-line-before)
```

Ref: https://github.com/stylelint/stylelint/issues/1573

jeddy3:
FYI, for someone wanting to implement this, a very similar rule, but for custom properties, can be found [here](https://github.com/stylelint/stylelint/tree/master/src/rules/custom-property-empty-line-before).

dryoma:
Added in #85 (at last :) )

dryoma:
Available in 1.4.1

PR #85, opened 2016-11-23
Dollar variable empty line before

dryoma:
Implements #53 , but with added `ignore` option and `after-dollar-variable` is used instead of `after-declaration`.


dryoma:
Some utils that were borrowed from stylelint (`beforeBlockString`, `blockString`) are used in their older state, without Flow definitions, to avoid the need to add es2017 preset as dependency.

dryoma:
Merging this. If anyone decides to have a look and find issues/suggest improvements, just make another PR or drop a line here and I'll make regular commit fixes.

Issue #82, opened 2016-11-07
New rule: same line or newline for `@else` statements


ntwb:
via https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#elseplacement

newline:
```css
@if {
  ...
}
@else {
  ...
}
```

same line:
```css
@if {
  ...
} @else {
  ...
}
```

dryoma:
Wow, we still don't have it :) Thanks.

kristerkari:
Yeah, definitely should have this.

dryoma:
Hmm, seems this rule won't be that easy.

Firstly, to be able to control `@else` placement at all, a user would need to set up ignore options for [`block-closing-brace-newline-after`](http://stylelint.io/user-guide/rules/block-closing-brace-newline-after/) like so: `ignoreAtRules: ["if", "else"]`. I would allow this rule to take effect, but would lead to:

```scss
@if ($f == 1) {
 ... 
} @else {} a { display: block; }
```
(notice the `a { display: block; }`). To prevent that we'll have to control how the closing brace of the last `@else` in a row behaves. The options I see:
1. Somehow read the `block-closing-brace-newline-after`'s config if present and apply it, or
1. Use our own option, specific to this rule, something like
```json
"at-else-newline-before": [
  "always", {
    "lastClosingBraceNewlineAfter": "always"
  }
]
```
---------
Secondly, the space before `@else`. [`block-closing-brace-space-after`](http://stylelint.io/user-guide/rules/block-closing-brace-space-after/) doesn't give any viable option, so we'll have to control it. The question is, should it be an option to this rule, or a standalone rule? `at-else-space-before` soulds stylelint-ish, but the whole rule for a single statement... I don't know, doesn't sound well.

And all that said, should this rule still be `at-else-newline-before` or `at-else-behavior`?

Any thoughts?

dryoma:
Hmm, seems this rule won't be that easy.

Firstly, to be able to control `@else` placement at all, a user would need to set up ignore options for [`block-closing-brace-newline-after`](http://stylelint.io/user-guide/rules/block-closing-brace-newline-after/) like so: `ignoreAtRules: ["if", "else"]`. I would allow this rule to take effect, but would lead to:

```scss
@if ($f == 1) {
 ... 
} @else {} a { display: block; }
```
(notice the `a { display: block; }`). To prevent that we'll have to control how the closing brace of the last `@else` in a row behaves. The options I see:
[...]

PR #87, opened 2016-11-28
If else newlines

dryoma:
Implements #82 

These rules with the corresponding options are added (the reasoning and examples are in the issue discussion):
```
"at-if-closing-brace-newline-after": "always-last-in-chain"
"at-if-closing-brace-space-after": "always-intermediate"|"never-intermediate"
"at-else-closing-brace-newline-after": "always-last-in-chain"
"at-else-closing-brace-space-after": "always-intermediate"|"never-intermediate"
"at-else-empty-line-before": "never"
```

I also added the /docs folder, for now it only has [examples/if-else.md](https://github.com/kristerkari/stylelint-scss/blob/if-else-newlines/docs/examples/if-else.md).

There are still cases I belive are unresolved:

- [ ] We can't force `@else if` to always be on the same line. Can be solved by the `never` option to [`at-rule-name-newline-after`](http://stylelint.io/user-guide/rules/at-rule-name-newline-after/) (I'm ok with pushing a PR).
- [x] Disallow an empty line before @else when `at-rule-empty-line-before` has `ignoreAtRules: [ "else" ]`.

@ntwb Please see if the functionality that is complete so far works for you, and if there are any other things we need done.

cc. @kristerkari @jeddy3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 - Added: `declaration-nested-properties-no-divided-groups` rule.
 - Added: `ignore: "local"|"global"` to the `dollar-variable-pattern` rule.
 - Added: `dollar-variable-empty-line-before` rule.
+- Added: `at-else-closing-brace-newline-after` rule.
+- Added: `at-else-closing-brace-space-after` rule.
+- Added: `at-if-closing-brace-newline-after` rule.
+- Added: `at-if-closing-brace-space-after` rule.
 
 # 1.3.4
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Added: `at-else-closing-brace-space-after` rule.
 - Added: `at-if-closing-brace-newline-after` rule.
 - Added: `at-if-closing-brace-space-after` rule.
+- Added: `at-else-empty-line-before` rule.
 
 # 1.3.4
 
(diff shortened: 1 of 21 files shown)


kristerkari:
Thanks for this, I'll try to find some time to go through the code tonight or tomorrow.

ntwb:
Super weird, I've not been getting notifications for this issue or PR, I'll try to look at it later today 👍

kristerkari:
I went through the code and looks good to me. 👍  Also, it probably makes sense to add "never" option to `at-rule-name-newline-after` if there is no specific reason why that rule can't have that option :)

dryoma:
Added `at-else-empty-line-before` rule so that we could pass this:
```scss
a { ... }

@if {
  // ...
}
@else {
  // ...
}
```

but not this:
```scss
@if {
  // ...
}

@else {
  // ...
}
```
(see the updated second example [here](https://github.com/kristerkari/stylelint-scss/blob/if-else-newlines/docs/examples/if-else.md))

dryoma:
Turns out that `at-rule-name-newline-after: "never"` will force the code to be 
```scss
@elseif () {}
```
due to how `whitespaceChedker` util works (`always-` options require either a space or a newline depending on a corresponding param, and `never-` options disallow whitespace altogether).

So for merging this as is for now. Will add a standalone rule for `@else` if users ask for it.

Issue #86, opened 2016-11-23
at-rule-no-unknown errors


neverfox:
I get at-rule-no-unknown errors for things like @extend, @mixin etc. with this plugin installed. I could be mistaken, but I assumed this plugin would cover those cases, but right now I have to use a custom ignore rule:

```
{
  "extends": [
    "stylelint-config-standard",
    "stylelint-config-css-modules"
  ],
  "plugins": [
    "stylelint-scss"
  ],
  "rules": {
    "at-rule-no-unknown": [ true, {
      ignoreAtRules: ['extend', 'at-root', 'debug', 'warn', 'error', 'if', 'else', 'for', 'each', 'while', 'mixin', 'include', 'content', 'return', 'function']
    }]
  }
}
```

dryoma:
The list of *known* rules is defined in [stylelint's code](https://github.com/stylelint/stylelint/blob/master/src/reference/keywordSets.js#L487). This project loads stylelint as a dependency, thus we can use its code in our code, but can not alter the rules from a standalone stylelint users define in their configs. ... or can we @davidtheclark ? Is it possible to give a plugin set the ability to redefine things like objects in keywordSets?

Alternatively something like `stylelint-scss/at-rule-no-unknown` can be used instead of the core rule; it would be the same except it would utilize our extended set of known rules. The problem would be we'll also have to keep up with the currently applicable standart atrules, so the first option I think is preferrable.

eriklharper:
@neverfox I'm in the same boat, but what I realized is that because `stylelint-scss` is a stylelint `plugin`, it doesn't automatically enable any rule for you because [that's how plugins work](http://stylelint.io/user-guide/configuration/#plugins).

What we _really_ need to make the linter "scss aware" is a scss lint config that we can tell stylelint to `extend` instead, the same way you're extending `stylelint-config-css-modules` above.

I did a quick search online for "stylelint scss config" but it isn't turning up anything, perhaps this is a perfect moment for me to attempt to write one myself!

eriklharper:
Here we go, there actually is one: https://github.com/bjankord/stylelint-config-sass-guidelines/blob/master/index.js.  See how he's using `stylelint-scss` as a plugin and then implementing `rules` on top of that plugin to make it scss aware.

neverfox:
@eriklharper Cool. Makes sense. Thanks for doing the leg work.

davidtheclark:
[...]

PR #92, opened 2016-12-07
Add rule: at-rule-no-unknown

dryoma:
Implements #86. Needed to consider SCSS-specific @-directives as known rules
while being able to disallow all the other non-standard unknown at-rules.

The only issue I couldn't resolve is how to import warning messages from core stylelint rules without hardcoding the path.

*Upd*: maybe postpone this until https://github.com/stylelint/stylelint/pull/2173 gets released and we're able to report from inside this rule (that would also serve the warning message issue from above).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Added: `declaration-nested-properties-no-divided-groups` rule.
 - Added: `ignore: "local"|"global"` to the `dollar-variable-pattern` rule.
 - Added: `dollar-variable-empty-line-before` rule.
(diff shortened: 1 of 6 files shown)


kristerkari:
@dryoma Sorry for checking this so late, but this looks good, and the stylelint PR that you mention has been merged. So maybe you'd want to rebase and then merge?

dryoma:
I do, and I will eventually (it needs some edits afiak)

kristerkari:
This was replaced by:
https://github.com/kristerkari/stylelint-scss/pull/127

Issue #64, opened 2016-07-28
TypeError: Cannot read property '1' of null


evilebottnawi:
```shell
TypeError: Cannot read property '1' of null
    at findCommentsInRaws (/path/to//node_modules/stylelint-scss/dist/utils/findCommentsInRaws.js:144:34)
    at /path/to//node_modules/stylelint-scss/dist/rules/double-slash-comment-whitespace-inside/index.js:18:50
    at /path/to//node_modules/stylelint/dist/postcssPlugin.js:134:46
    at Array.forEach (native)
    at /path/to//node_modules/stylelint/dist/postcssPlugin.js:113:33
```

Got this error on:
```scss
/*!
 *  Font Awesome 4.6.3 by @davegandy - http://fontawesome.io - @fontawesome
 *  License - http://fontawesome.io/license (Font: SIL OFL 1.1, CSS: MIT License)
 */
```

and

```scss
/* Example with animate.css
@import "animate.css/source/_base.css";
@import "animate.css/source/fading_entrances/fadeIn.css";
*/
```
Got this error on rules:
- `partial-no-import`
- `double-slash-comment-empty-line-before`
- `double-slash-comment-inline`
- `double-slash-comment-whitespace-inside`
- `operator-no-newline-after`
- `operator-no-newline-before`
- `operator-no-unspaced`

dryoma:
Wow, thanks.

dryoma:
Please update to 1.3.1 and see if it is fixed for you.

evilebottnawi:
@dryoma thanks! I'm going to try now

evilebottnawi:
@dryoma 
The error still occurs:
```shell
TypeError: Cannot read property '1' of null
    at findCommentsInRaws (/home/itgalaxy/web/dev0.cms.dev.itgalaxy.company/public_html/wp-content/themes/insteria/node_modules/stylelint-scss/dist/utils/findCommentsInRaws.js:76:123)
    at /home/itgalaxy/web/dev0.cms.dev.itgalaxy.company/public_html/wp-content/themes/insteria/node_modules/stylelint-scss/dist/rules/double-slash-comment-whitespace-inside/index.js:18:50
    at /home/itgalaxy/web/dev0.cms.dev.itgalaxy.company/public_html/wp-content/themes/insteria/node_modules/stylelint/dist/postcssPlugin.js:134:46
    at Array.forEach (native)
[...]

PR #96, opened 2016-12-26
allow backslash as a part of valid function name

whizark:
[#64 TypeError: Cannot read property '1' of null](https://github.com/kristerkari/stylelint-scss/issues/64) still occurs under certain conditions.

I have struggled with the errors because they had seemed to be fixed. :fearful:
Finally, I noticed that the issue comes from function or mixin name that includes `\` (backslash).

It would be nice stylelint-scss accepts `\` as a part of valid function/mixin name, then the errors won't occur anymore.

ref. sass/libsass/pull/561

---
The following is a scss file that reproduces the error.

test.scss
```scss
$foo\bar: 1;

@function baz\qux() {
}

@mixin quux\corge() {
}
```

## Environment
* Ubuntu 16.04 LTS
* stylelint 7.7.0
* stylelint-scss 1.4.1

---
**Added** (2016/12/27)
This PR needs more modification.
`\` in [at]rule also produces incorrect column poisitions. It seems to be [postcss-scss](https://github.com/postcss/postcss-scss)'s issue rather than stylelint-scss's.

### DIFF
--- a/src/utils/__tests__/findCommentsInRaws.js
+++ b/src/utils/__tests__/findCommentsInRaws.js
@@ -265,3 +265,51 @@ test("CSS comment, Windows newlines", t => {
     })
     .catch(logError)
 })
+
+test("//-comment, backslash in identifier", t => {
+  t.plan(4)
+
+  postcss()
+    .process(`
+      $foo\bar: 1; // variable
+
+      @function baz\qux() // function
+      {}
+       
+      @mixin quux\corge() // mixin
+      {}
+    `, { syntax: scss })
+    .then(result => {
+      const css = result.root.source.input.css
+      const comments = findCommentsInRaws(css)
+      t.equal(comments.length, 3)
+      t.deepEqual(comments[0].source, { start: 20, end: 30 })
+      t.deepEqual(comments[1].source, { start: 61, end: 71 })
+      t.deepEqual(comments[2].source, { start: 118, end: 125 })
+    })
+    .catch(logError)
+})
+
+test("CSS comment, backslash in identifier", t => {
+  t.plan(4)
+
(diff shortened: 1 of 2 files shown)


dryoma:
Good point, thanks. Could you please add some [test cases](https://github.com/whizark/stylelint-scss/blob/fc8e67a5e288caffdeb5ab331653d2eb820ddec9/src/utils/__tests__/findCommentsInRaws.js) for that change?

whizark:
After that, I also noticed there is another issue that produces incorrect comment start/end position.
`\` character skips one column position and it probably comes from postcss-scss.

* https://github.com/postcss/postcss-scss/blob/master/lib/scss-tokenize.es6#L26
* https://github.com/postcss/postcss-scss/blob/master/lib/scss-tokenize.es6#L216

I wrote some tests but they failed as expected... so this PR needs more modification or fix `postcss-scss`'s issue.
I'll try to find a way out to fix them later but I shouldn't use `/` in function/mixin name as it is an unusual character for now.

kristerkari:
@whizark Are you still working on this? Do you need help with anything?

whizark:
@kristerkari Hi, I actually try not to use unusual `\` now... so would you close this PR as needed?

I probably should open/send an issue/PR to `postcss` because it seems that the issue isn't only for `stylelint-scss` and `postcss-scss` but also for `postcss` itself, .
https://github.com/postcss/postcss/blob/master/lib/tokenize.es6#L21

https://github.com/postcss/postcss/issues/923 seems to be a related issue.

kristerkari:
Closing this as it is something that is probably not going to be supported by postcss parser.

PR #99, opened 2017-02-08
Fix at-mixin-argumentless-call-parentheses messages

YodaDaCoda:


dryoma:
Hi @YodaDaCoda 

Good call, thanks! Just FYI, stylelint (and thus we) has a convention for error messages, in this case they should be:
```
expected: "Expected parentheses in @mixin call",
rejected: "Unexpected parentheses in argumentless @mixin call"
```
Added proper edits with my commit, and also added a mixin names to the messages.

YodaDaCoda:
@dryoma Thanks for the info! I'll make sure I keep that in mind for any future PRs.

PR #108, opened 2017-03-30
Change ruleName in at-mixin-argumentless-call-parentheses to match na…

### LINKED ISSUE #106: at-mixin-argumentless-call-parentheses rule is showing up severity: "ignore"
Hey,
I have seen that such a bug was already resolved in version 1.1.0 but it still occurs in at-mixin-argumentless-call-parentheses rule.

danrr:
…me of the rule

at-mixin-argumentless-call-parentheses had an extra no- in its ruleName that caused the rule to fail
with an ignore message rather than an error.

Fixes #106


dryoma:
Thanks!

PR #110, opened 2017-04-09
at-if-closing-brace-newline-after: support @elseif

### LINKED ISSUE #107: at-if-closing-brace-newline-after only checks for else and not elseif
If you use the `at-if-closing-brace-newline-after` rule with `always-last-in-chain` the following code generates an error:
```scss
@if $min and $max {
	$range: '(#{$min} - #{$max})';
} @elseif $max {
	$range: '(< #{$max})';
} @else {
	$range: '(> #{$min})';
}
```

The happens because elseif is not checked. Although not documented in the [sass-documentation](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#_8) `@elseif` is supported in scss.

The code to update is here:
https://github.com/kristerkari/stylelint-scss/blob/483052/src/rules/at-if-closing-brace-newline-after/index.js#L62

kristerkari:
Add support for `@elseif` for `at-if-closing-brace-newline-after` rule. [It works in Sass](https://github.com/sass/libsass/issues/1060) (but it's not documented).

fixes #107

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# HEAD
+
+- Fixed: `at-if-closing-brace-newline-after`: support `@elseif`.
+
 # 1.4.3
 
 - Fixed: `at-mixin-no-argumentless-call-parentheses` messages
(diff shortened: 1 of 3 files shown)


kristerkari:
ping @rvock @dryoma

dryoma:
Looks good to me.

kristerkari:
Sure I can do that.

PR #150, opened 2017-07-28
Feature: migrate to `stylelint@8.0.0`.

evilebottnawi:
Notes:
1. Support npm@5 (add `package-lock.json`)
2. We change logic `ignore*` (see https://github.com/stylelint/stylelint/pull/2709)
3. Use `stylelint` build-in `isStandardSyntax*` (less code - best support)
4. Remove `replace` package (is not used anywhere)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,5758 @@
+{
+  "name": "stylelint-scss",
+  "version": "1.5.1",
+  "lockfileVersion": 1,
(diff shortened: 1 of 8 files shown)


kristerkari:
Thanks for taking over this!

I'll release new versions tomorrow. First a patch with the current fixes in master and then a major version with these changes. :+1:

Issue #112, opened 2017-04-12
selector-no-redundant-nesting-selector Doesn't Error On Multi-part Selectors


epmatsw:
```
.a {
    & .b,
    & .c {

    }
}
```

An error is correctly reported on the `& .b` line, but no error is reported on the `& .c` line. Seems like all of the pieces of the selector need to be examined by this rule.

PR #155, opened 2017-07-31
Fixed: `selector-no-redundant-nesting-selector` now handle multiple nested selectors.

evilebottnawi:
Fix: https://github.com/kristerkari/stylelint-scss/issues/112

### DIFF
--- a/src/rules/selector-no-redundant-nesting-selector/index.js
+++ b/src/rules/selector-no-redundant-nesting-selector/index.js
@@ -1,5 +1,5 @@
 import { utils } from "stylelint";
-import { namespace } from "../../utils";
+import { namespace, parseSelector } from "../../utils";
 
 export const ruleName = namespace("selector-no-redundant-nesting-selector");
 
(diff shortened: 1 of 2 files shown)


src/rules/selector-no-redundant-nesting-selector/index.js
```diff
@@ -15,18 +15,43 @@ export default function(actual) {
...
-      const regex = /^&\s*[>+~ ]\s*[^>+~ {&]+/;
-
-      if (selector === "&" || regex.test(selector)) {
-        utils.report({
-          ruleName,
-          result,
-          node: rule,
-          message: messages.rejected
+      parseSelector(rule.selector, result, rule, fullSelector => {
+        // "Ampersand followed by a combinator followed by non-combinator non-ampersand and not the selector end"
```
kristerkari: You can remove this comment now as it was for the regular expression

kristerkari:
Awesome work! Thanks a lot

src/rules/selector-no-redundant-nesting-selector/index.js
```diff
@@ -15,18 +15,43 @@ export default function(actual) {
...
+        fullSelector.walkNesting(node => {
+          const prev = node.prev();
+
+          if (prev) {
+            return;
+          }
+
+          const next = node.next();
+
+          if (!next && node.parent.parent.nodes.length > 1) {
```
kristerkari: Is there a chance that this could throw? For example if `node.parent.parent` would be `undefined`?

src/rules/selector-no-redundant-nesting-selector/index.js
evilebottnawi: @kristerkari i think be good leave him for clarify what we do for other contributors

src/rules/selector-no-redundant-nesting-selector/index.js
evilebottnawi: @kristerkari any nested element have `parent` and any `parent` have `parent` as full selector, but in next version behavior in theory can be changed, we can add more checks to avoid this situation.

kristerkari: Ok, so it's good enough if it can't throw an error with the current setup

Issue #181, opened 2017-11-13
Rule Idea: at-mixin-named-arguments and at-function-named-arguments


diego-codes:
## Context
I looked around the project for a mention to this type of rule and didn't find anything, so I apologize if this is a duplicate.

Sass has a really useful feature that allows developers to call mixins and functions with named parameters. It makes otherwise confusing calls with "mystery" arguments more readable and easy to update. I didn't write this [article](http://blog.ricardofilipe.com/post/object-arguments-in-sass), but I agree with this approach. I am advocating in my main project for contributors to adhere to this writing style. I would love it if stylelint-scss checked for this.

## Rules
I believe the two rules should work the same way. The options for both rules could be: `always`, `never`, `always-multiple-arguments`.

### `always`
Bad:
```scss
.foo {
  margin-top: size(10);
}

// or
.foo {
  font-size: em-scale(20, 16);
}

// or 
.foo {
  @include animation(250ms, 100ms, infinite);
} 

// or
.foo {
  @include my-mixin($value: 20, 'bar', $color: #FFF);
}
```

Good:
```scss
.foo {
  font-size: em-scale($size: 20, $base: 16);
}

// or 
.foo {
  @include animation($duration: 250ms, $delay: 100ms, $direction: infinite);
}
```

### `never`
Bad:
```scss
.foo {
  font-size: em-scale($size: 20, $base: 16);
}

// or 
.foo {
  @include animation($duration: 250ms, $delay: 100ms, $direction: infinite);
}

// or
.foo {
  @include my-mixin($value: 20, 'bar', $color: #FFF);
}
```

Good:
```scss
.foo {
  margin-top: size(10);
}

// or
.foo {
  font-size: em-scale(20, 16);
}

// or 
.foo {
  @include animation(250ms, 100ms, infinite);
} 
```

### `always-multiple-arguments`
Bad:
```scss
.foo {
  font-size: em-scale(20, 16);
}

// or 
.foo {
  @include animation(250ms, 100ms, infinite);
}

// or
.foo {
  @include my-mixin($value: 20, 'bar', $color: #FFF);
}
```

Good:
```scss
.foo {
  margin-top: size(10);
}

// or
.foo {
  font-size: em-scale($size: 20, $base: 16);
}

// or 
.foo {
  @include animation($duration: 250ms, $delay: 100ms, $direction: infinite);
}
```

kristerkari:
Thanks! This seems like a really good idea for a rule. :+1:

[...]

PR #182, opened 2017-11-21
New rule: mixin and function named arguments

diego-codes:
Implements #181 

Two separate rules: `at-function-named-parameters` and `at-mixin-named-parameters` that implement the similar logic.


kristerkari:
Thanks! I'll try to get this reviewed asap :)

src/rules/at-function-named-arguments/README.md
```diff
@@ -0,0 +1,115 @@
+# at-function-named-arguments
+
+Require named parameters in SCSS function call rule.
+
+```scss
+animation: animation($duration: 250ms) {
+//                 ↑
```
kristerkari: Small thing, but this arrow should point to the `$`.

src/rules/at-function-named-arguments/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+```scss
+.foo {
+  border: reset($value: 20, 'bar', $color: #FFF);
+}
+```
+
+```scss
+.foo {
+  animation: animation($duration: 250ms);
```
kristerkari: Why is this considered a warning when the rule is set to `always-multiple-arguments`?

src/rules/at-mixin-named-arguments/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+```scss
+.foo {
+  @include reset($value: 20, 'bar', $color: #FFF);
+}
+```
+
+```scss
+.foo {
+  @include animation($duration: 250ms);
```
kristerkari: Same question as with Sass functions: should this be a warning if the rule is set to `always-multiple-arguments´?

kristerkari:
Thanks @diego-codes! Sorry that it took me so long to review this, but I've been quite busy this week.

I added a couple of comments for things that I was unsure of. If you could check them out and after that I can merge this PR and release a new version.

Good job :+1:

src/rules/at-function-named-arguments/README.md
diego-codes: That’s is a good point. Would `only-multiple-arguments` make sense?

kristerkari: I did a search in `stylelint` repo, but I didn't see `only-` prefix used in any of the rules, so maybe we should also try to keep things consistent.

I think that we can keep the name as it is right now, or then do something like `"always"` with a secondary option `{ except: ["single-argument"] }`, but I think that that might just be more confusing.

I'll need to sleep the night first to be able to think clearly :)

diego-codes: Hey, sorry I'm just getting back to this. I agree it's important to keep things consistent. 

If we add the secondary `{ "except": ["single-argument"] }` option, would using it along with `"never"` mean that single arguments have to be named?

Looking at `stylelint`'s guidelines another approach can be `{ "ignoreSingleArgument": true }` which basically would avoid checking single arguments for either `"never"` or `"always"`. I'm not sure that I like this either, but it's more clear then `"except"` in my opinion.

What do you think?

kristerkari: Using `"ignore"` could work, so how about `{ "ignore": ["single-argument"] }`? That style would also be extendable in the future.

diego-codes: That sounds good. I’ll make this change.

diego-codes:
Hey @kristerkari, I removed the `"always-multiple-arguments"` primary option and added the `"ignore: ["single-argument"]"` secondary option based on our discussion. The number of tests basically doubled which may be excessive...

Let me know what you think!

kristerkari:
@diego-codes  Awesome! I would not be too worried about having a lot of tests. If you want to reduce duplication, you can use `.concat` like this rule does https://github.com/kristerkari/stylelint-scss/blob/master/src/rules/double-slash-comment-empty-line-before/__tests__/index.js#L135-L148

Let me know if you still want to make any changes. If no changes are needed, I can merge this 👍

src/rules/at-function-named-arguments/index.js
```diff
@@ -0,0 +1,133 @@
...
+const isScssVarRegExp = /^\$\S*/;
+
+export default function(expectation, options) {
+  return function(root, result) {
+    const validOptions = utils.validateOptions(
+      result,
+      ruleName,
+      {
+        actual: expectation,
+        possible: ["always", "never", "always-multiple-arguments"]
```
kristerkari: ` "always-multiple-arguments"` should be remove from here.

src/rules/at-mixin-named-arguments/index.js
```diff
@@ -0,0 +1,112 @@
...
+const isScssVarRegExp = /^\$\S*/;
+
+export default function(expectation, options) {
+  return function(root, result) {
+    const validOptions = utils.validateOptions(
+      result,
+      ruleName,
+      {
+        actual: expectation,
+        possible: ["always", "never", "always-multiple-arguments"]
```
kristerkari: ` "always-multiple-arguments"` should be remove from here too.

src/rules/at-function-named-arguments/index.js
```diff
@@ -0,0 +1,133 @@
...
+    "Unexpected a named parameter in single argument function call"
+});
+
+const cssFunctions = [
+  "attr",
+  "calc",
+  "linear-gradient",
+  "radial-gradient",
+  "repeating-linear-gradient",
+  "repeating-radial-gradient"
```
kristerkari: Oh, btw. this function list seems incomplete, here's the full one:
https://github.com/EtaiG/css-functions-list

kristerkari: and maybe you could also add a test that tests a CSS function that uses camelCase, for example `translateZ`.

diego-codes: Good call!

diego-codes:
@kristerkari, I made the changes you requested. I think I'm okay with the tests as they are. I just added some to tests for CSS functions.

src/rules/at-function-named-arguments/index.js
```diff
@@ -0,0 +1,125 @@
...
+  isNativeCssFunction
+} from "../../utils";
+import valueParser from "postcss-value-parser";
+
+export const ruleName = namespace("at-function-named-arguments");
+
+export const messages = utils.ruleMessages(ruleName, {
+  expected: "Expected a named parameter to be used in function call",
+  rejected: "Unexpected a named parameter in function call",
+  rejectedSingle:
```
kristerkari: I had another look at the code and I noticed that `rejectedSingle` does not seem to be used anywhere. It's also in the other rule.

src/rules/at-function-named-arguments/__tests__/index.js
```diff
@@ -0,0 +1,852 @@
...
+  reject: [
+    {
+      code: `
+      .b {
+        border: reset($value: 40px);
+      }
+      `,
+      line: 3,
+      column: 9,
+      message: messages.unexpected,
```
kristerkari: Here and in the lines below the message should be `messages.rejected` as there is no `messages.unexpected` defined.

This also happens in the the tests for the other rule.

kristerkari:
Great work @diego-codes! I found still a couple of really small things, if you could fix those, then I can merge this and release a new version tomorrow. :)

diego-codes:
@kristerkari, sorry about that. That was really sloppy editing. Thanks for catching it. Let me know if you see anything else. Otherwise, I think it's good to go!

Issue #198, opened 2018-01-31
Rule to check !default flag


freezy-sk:
https://github.com/twbs/bootstrap/issues/20717 require linting that all variables have !default flag
What about name `dollar-variable-default`?
Primary option: `true`
Secondary options: `ignore: ["local"]` to ignore local variables and check only global

kristerkari:
Thanks! Sounds good :)

PR #199, opened 2018-02-01
dollar-variable-default rule #198

freezy-sk:
WIP #198

### DIFF
--- a/src/rules/dollar-variable-default/__tests__/index.js
+++ b/src/rules/dollar-variable-default/__tests__/index.js
@@ -28,17 +28,35 @@ testRule(rule, {
     {
       code: "$var: 10px",
       message: messages.expected("$var"),
-      description: "Global variable without !default"
+      description: "Global variable without !default",
+      line: 1,
+      column: 1
     },
     {
-      code: "a { $var: 10px }",
-      message: messages.expected("$var"),
-      description: "Local variable without !default"
+      code: `
+        $global: 10px !default;
+        a {
+          $local-var: 10px;
+        }
+      `,
+      message: messages.expected("$local-var"),
+      description: "Local variable without !default",
+      line: 4,
+      column: 11
     },
     {
-      code: ".class { a { $var: 10px } }",
-      message: messages.expected("$var"),
-      description: "Nested local variable without !default"
+      code: `
+        $global: 10px !default;
+        .class {
+          a {
+            $nested-var: 10px;
+          }
+        }
+      `,
+      message: messages.expected("$nested-var"),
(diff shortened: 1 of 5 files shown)


kristerkari:
Ping me when you want me to review this 👍

freezy-sk:
@kristerkari I've added readme file for rule so you can review it right now :)

kristerkari:
Thanks a lot, great job! 

The code looks good and simple, only some small things came to mind related to the unit tests:
- You could add the same test cases using nested selectors (e.g. `.class { a { $var: 10px !default  } }`) to both `accept` and `reject` for both `ignore` option and the rule with default settings. This way we can make sure that the rule also works when the variables are nested.
- You could add a simple `a { color: blue  }` test case to `accept` just to make sure that no one removes the check for declaration starting with `$` in the future.

Also, if you don't mind, you could add this rule to the main README under list of rules: https://github.com/kristerkari/stylelint-scss#list-of-rules

src/rules/dollar-variable-default/__tests__/index.js
```diff
@@ -0,0 +1,58 @@
...
+  ],
+
+  reject: [
+    {
+      code: "$var: 10px",
+      message: messages.expected,
+      description: "Global variable without !default"
+    },
+    {
+      code: "a { $var: 10px }",
```
kristerkari: oh yeah, could you add `line` and `column` numbers to the rejected messages just to make sure that the error is shown at the correct place (`$`-sign)?

Like here:
https://github.com/kristerkari/stylelint-scss/blob/master/src/rules/double-slash-comment-inline/__tests__/index.js#L43-L52

kristerkari:
I don't know if you noticed this comment, as Github decided to hide it:
https://github.com/kristerkari/stylelint-scss/pull/199/files/5af02874200008146b84c1b2e25ced82c54432d4#r165815199

Before merging this, I need to revert the changes that drop support for NodeJS 4.x. That's because stylelint has not released a major version yet, and I don't want to release a major version before they do. By reverting the changes, I can release this as part of a minor version release instead.

freezy-sk:
@kristerkari I finished incorporating all of your hints right now

freezy-sk:
@kristerkari you were so fast with that review :D

kristerkari:
Thanks! I'll merge & release new version after I've reverted a couple of commits. 👍

freezy-sk:
thank you :)

kristerkari:
https://github.com/kristerkari/stylelint-scss/releases/tag/2.3.0 🎉

PR #217, opened 2018-03-03
dollar-variable-colon-newline-after: always allow multiline vars with parens

### LINKED ISSUE #172: Rule Improvement for maps - dollar-variable-colon-newline-after
Currently enabling **dollar-variable-colon-newline-after** errors for:

````sass
$map: (
  foo: value,
  bar: value,
);
````

Though it's nice and readable.

To fix it I'm forced to write:

````sass
$map: 
  (
    foo: value,
    bar: value,
  );
````

kristerkari:
See discussion in: https://github.com/kristerkari/stylelint-scss/issues/172#issuecomment-369716880

Fixes #172

Please review this @levacic @sindresorhus @AndyOGo

### DIFF
--- a/src/rules/dollar-variable-colon-newline-after/__tests__/index.js
+++ b/src/rules/dollar-variable-colon-newline-after/__tests__/index.js
@@ -205,6 +205,43 @@ testRule(rule, {
       `,
       description:
         "always-multi-line: should allow multiline variable using newline after colon"
+    },
+    {
+      code: `
+      $map: (
+        foo: 1,
+        bar: 2,
+      );
+      `,
+      description:
+        "always-multi-line: should allow using Sass map without a newline"
+    },
+    {
+      code: `
+      $var: (
+        1 +
+        2 +
+        3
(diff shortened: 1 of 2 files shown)


levacic:
I haven't tested the code but it looks like it's doing the right thing and I see the tests are passing :+1: 

As mentioned in the original issue though, I'm not sure if everyone would expect the same behavior with `always-multi-line` by default, or whether this should perhaps be part of an `except` or `ignore` option. But this does conform to code that my team finds readable, so we definitely appreciate that we can use this linter now.

kristerkari:
> As mentioned in the original issue though, I'm not sure if everyone would expect the same behavior with always-multi-line by default, or whether this should perhaps be part of an except or ignore option.

It's a bit hard to decide on what to do with changes like this one, but as I already changed the behavior for `always` option, I don't think that this change will upset a lot of people. 

At least there are me + 3 persons currently who don't seem to have anything against doing this change.

PR #225, opened 2018-03-08
Add autofix for double-slash-comment-empty-line-before

kristerkari:


src/rules/double-slash-comment-empty-line-before/index.js
```diff
@@ -90,6 +90,25 @@ export default function(expectation, options) {
         return;
       }
 
+      if (context.fix) {
+        if (expectEmptyLineBefore && !hasEmptyLineBefore) {
+          if (/^\n\s*?$/.test(before)) {
+            comment.raws.before = before.replace(/^\n/, "\n\n");
+          } else if (/^\r\n\s*?$/.test(before)) {
+            comment.raws.before = before.replace(/^\r\n/, "\r\n\r\n");
```
OriR: Won't it be better to just do something like this:
```js
if (new RegExp(`/^${context.newLine}\\s*?$/`).test(before)) {
comment.raws.before = before.replace(new RegExp(`/^${context.newLine}/`), context.newLine + context.newLine);
}
```
?

That way you can extract it to a function and use it in both cases:
```js
const fix = (comment, match, replace) => {
  if (new RegExp(`/^${match}\\s*?$/`).test(before)) {
    comment.raws.before = before.replace(new RegExp(`/^${match}/`), replace);
  }
}
```
Although now that I think about it, you don't really need to test if the `before` contains the new line, you know it does, but you're not sure about the line endings which is indicated by `context.newLine`. So it can be simplified to this:
```js
const fix = (comment, match, replace) => {
  comment.raws.before = before.replace(new RegExp(`/^${match}/`), replace);
}
```
and you'd call it here with:
```js
fix(comment, context.newLine, context.newLine + context.newLine);
```
And below with:
```js
fix(comment, context.newLine + context.newLine, context.newLine);
```

kristerkari: Good points, I'll try that out and let's see if it works!

kristerkari: @OriR I tried your `fix` function, but the tests start failing when I used it. 

Maybe you would like to implement a fix against this branch and make another PR?

OriR: I'll give it a look 👍

OriR: Created #230 
My RegExp game is not that strong in Github comments 😜

kristerkari:
Let's use #230 instead of this 👍

PR #226, opened 2018-03-08
Add autofix for dollar-variable-empty-line-before

kristerkari:


src/rules/dollar-variable-empty-line-before/__tests__/index.js
```diff
@@ -334,6 +367,11 @@ testRule(rule, {
     {
       code: `a {
       width: 1; $var2: 2;
+    }`,
+      fixed: `a {
+      width: 1;
+
+$var2: 2;
```
kristerkari: This is again a bit problematic, the variable is not indented when a newline is added.

src/rules/dollar-variable-empty-line-before/index.js
```diff
@@ -93,10 +93,41 @@ export default function(expectation, options) {
...
 
+      if (context.fix) {
+        if (expectHasEmptyLineBefore && !hasEmptyLine(before)) {
+          if (optionsHaveException(options, "first-nested")) {
+            decl.raws.before = before.replace(
+              /^\s+/,
+              context.newline + context.newline
+            );
+          }
+          if (/^\n\s*?$/.test(before) || /^\r\n\s*?$/.test(before)) {
```
kristerkari: This whole regex mess needs a similar `fix` function as in #230

src/rules/dollar-variable-empty-line-before/__tests__/index.js
OriR: See my [comment](https://github.com/kristerkari/stylelint-scss/pull/231#discussion_r173961660) regarding the indentation problem in autofixing - we should take a similar approach here.

kristerkari: @OriR I have been working on other stuff the for the last week or so, but I have now added `disableFix` option to this rule.

Would be nice to get the fixes in and release a new version :)

OriR: Yeah, I didn't have much time going over the PRs as well.

I'll try and implement it as soon as possible 👍

kristerkari: Don't worry, there's no hurry :) I'm just anxious with this stuff

OriR: So the `disableFix` should appear only here and `dollar-variable-colon-newline-after` since they're the "debatable" rules?
Is there any other rule?

kristerkari: I think that it's only those 2 rules with indentation issues.

PR #229, opened 2018-03-10
Add autofix for at-if/else-closing-brace-newline-after

OriR:


src/rules/at-else-closing-brace-newline-after/__tests__/index.js
```diff
@@ -80,6 +89,11 @@ testRule(rule, {
 
       } @else if { }
       @else { }
+    }`,
+      fixed: `a {
+      @if ($x == 1) {
+
+      } @else if { }@else { }
```
OriR: Not sure this fix makes sense 😕

src/rules/at-else-closing-brace-newline-after/__tests__/index.js
```diff
@@ -92,6 +106,11 @@ testRule(rule, {
       } @else { }
 
       @else { }
+    }`,
+      fixed: `a {
+      @if ($x == 1) {
+
+      } @else { }@else { }
```
OriR: Same here...

src/rules/at-else-closing-brace-newline-after/__tests__/index.js
kristerkari: In my opinion `} @else if { } @else { }` would look better :)

OriR: Makes sense, that's the same approach as in #221 👍

kristerkari:
Looks like you did some indentation changes in a0c2011 that broke the tests.

OriR:
@kristerkari Fixed it 👍 

I think we're good to go on all auto-fix PRs, right?

kristerkari:
Hmm... I'm thinking that maybe this rule should also have `disableFix` as it adds a newline without indentation. What do you think @OriR ?

OriR:
Yeah, on second look it looks odd.
Will add it now.

PR #230, opened 2018-03-10
Add autofix for double-slash-comment-empty-line-before

OriR:
replaces #225 

Because RegExp construction from string is a mess 🤷‍♂️

### DIFF
--- a/src/rules/double-slash-comment-empty-line-before/index.js
+++ b/src/rules/double-slash-comment-empty-line-before/index.js
@@ -15,7 +15,7 @@ export const messages = utils.ruleMessages(ruleName, {
 
 const stylelintCommandPrefix = "stylelint-";
 
-export default function(expectation, options) {
+export default function(expectation, options, context) {
   return (root, result) => {
     const validOptions = utils.validateOptions(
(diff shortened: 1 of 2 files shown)


src/rules/double-slash-comment-empty-line-before/index.js
```diff
@@ -90,6 +90,28 @@ export default function(expectation, options) {
         return;
       }
 
+      if (context.fix) {
+        const fix = (comment, match, replace) => {
```
kristerkari: This much cleaner than my PR!

One thing caught my eye though. The fix function is now inside the `root.walkComments` loop.  I think that it would be nicer to move that function outside the loop so that it's not re-declared on every loop:

```js
    const fix = (comment, match, replace) => {
      const escapedMatch = match.replace(
        /(\r)?\n/g,
        (_, r) => (r ? "\\r\\n" : "\\n")
      );
      comment.raws.before = comment.raws.before.replace(
        new RegExp(`^${escapedMatch}`),
        replace
      );
    };

    root.walkComments(comment => {
```

PR #231, opened 2018-03-10
Add autofix for dollar-variable-colon-newline-after

OriR:


src/rules/dollar-variable-colon-newline-after/index.js
```diff
@@ -60,6 +60,21 @@ export default function(expectation) {
           index: indexToCheck,
           lineCheckStr: decl.value,
           err: m => {
+            if (context.fix) {
+              const nextLinePrefix =
+                expectation === "always"
+                  ? decl.raws.before.replace(context.newline, "")
```
OriR: Not sure this is the right approach as the value will have the same indentation as the variable in the case of `always`.
I did it like this because I don't think I can get the indentation value from `stylelint` 😕 

Should we provide a `dontFix` option for this rule since this fix is pretty opinionated?

kristerkari: Yeah, autofixing in rules that can have indentation is a bit tricky. :/ Should we do similar indentation as the multiline `box-shadow` examples have?

> Should we provide a dontFix option for this rule since this fix is pretty opinionated?

I guess that makes sense.

OriR: I submitted an [issue](https://github.com/stylelint/stylelint/issues/3210) with stylelint to check if there's a way to perform autofixing that includes indentation - the answer was no. 
We shouldn't handle indentation, just fix the characters that we need, the rest is supposed to be handled by `stylelint` itself or other rules.

I'm still in favor of adding `dontFix` just in case it might cause problems.

kristerkari: I had a look at the stylelint issue and jeddy3's explanation makes sense. So, let's not worry about the missing indentation at all.

> I'm still in favor of adding dontFix just in case it might cause problems.

I'm not sure if we really need that, if stylelint's philosophy is: "Every rule is standalone and independent of other rules. So, a rule should just fix problems to make itself valid, and not worry about introducing violations for other rules e.g."

OriR: Technically we don't need this but, I think that's the reason [`stylelint-order`](https://github.com/hudochenkov/stylelint-order#autofixing) added the `disableFix` flag.

But that's your call, I'm just laying out the facts :)

kristerkari: Now that I have thought about it, it would probably be a good idea to have that flag when there is a big project that wants to use `--fix` but does not want to have tons wrongly indented variables.

So let's do it @OriR  👍

kristerkari: I have implemented `disableFix` option here:
https://github.com/kristerkari/stylelint-scss/pull/226/commits/57c22a448f8108b45d66a8d7a855ea34cb439342

kristerkari: @OriR as there are only a couple of small things missing from the autofixing, do you want me to help and add those changes?

OriR: Sorry about that... just found the time to continue the work.

Will update the PRs in the next couple of hours 👍

OriR:
@kristerkari Added `disableFix` :)

PR #248, opened 2018-06-06
fix: compatibility with non css syntaxes

evilebottnawi:
In stylelint we starting support non css syntax, i.e. `postcss-html`(vue and other html based syntaxes)/`postcss-markdown`/`postcss-css-in-js` to allow linting styles in these formats. Some rules require modification. After release `stylelint@9.3.0` it is possible to add tests on this changes.

Ref: https://github.com/stylelint/stylelint/blob/94174973b2bdb397830deae804ca2f5d46b19ab4/docs/user-guide/css-processors.md
Same PR in stylelint: https://github.com/stylelint/stylelint/pull/3367

Briefly: we migrate on `postcss-syntax` what change `postcss-*` parser based on extension and support non css syntaxes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,23 +5,23 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0-beta.49",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.49.tgz",
-      "integrity": "sha1-vs2AVIJzREDJ0TfkbXc0DmTX9Rs=",
+      "version": "7.0.0-beta.51",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.51.tgz",
+      "integrity": "sha1-vXHZsZKvl435FYKdOdQJRFZDmgw=",
       "dev": true,
       "requires": {
-        "@babel/highlight": "7.0.0-beta.49"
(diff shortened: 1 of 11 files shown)


kristerkari:
Thanks for this! 

I'll have a look at the links that you posted, but in the meantime one question came to my mind: guessing from the stylelint version, this should not be a breaking change, right?

evilebottnawi:
@kristerkari yep, but it is new feature and can be changes in 9 branch, between use own util helper for 9 version

kristerkari:
@evilebottnawi Is that what the `eachRoot` helper is for?

evilebottnawi:
@kristerkari yep, all non css syntaxes have `document` node with multiple `root` nodes

kristerkari:
Nice!

> After release stylelint@9.3.0 it is possible to add tests on this changes.

So is it possible to test these changes now after the release?

The changes don't seem that big anyway, so I can release a new version when everything is ready.

evilebottnawi:
@kristerkari let's do this

kristerkari:
Great! I'll release a new version after CI is green. by the way @evilebottnawi: I see that the same PR was marked as bugfixes to stylelint's changelog. Is this a new feature or simply a bugfix for something that you could already do?

evilebottnawi:
@kristerkari bugfix, looks some coverage decreased :confused:

kristerkari:
https://github.com/kristerkari/stylelint-scss/releases/tag/3.1.2

PR #255, opened 2018-06-16
fix: handle non css syntaxes

evilebottnawi:
Find another little problem, need patch release :smile:


kristerkari:
Thanks!

kristerkari:
https://github.com/kristerkari/stylelint-scss/releases/tag/3.1.3

PR #271, opened 2018-09-23
Fix crash when at-function-named-arguments encounters trailing commas in function calls

thibaudcolas:
When `at-function-named-arguments` is enabled, stylelint crashes when the rule attempts to parse function calls that have a trailing comma (has been [part of Sass since ± 2016](https://github.com/sass/libsass/issues/2070)). For example:

```scss
.b {
  border: reset(
    $value: 40px,
  );
}
```

This will cause the following error:

```
    TypeError: Cannot read property 'key' of undefined

       98 |             case "always": {
    >  99 |               if (arg.key && isScssVarRegExp.test(arg.key)) {
          |                                                    ^
      100 |                 return;
      101 |               }

      at src/rules/at-function-named-arguments/index.js:99:52
      [...]
```

---

The error is caused by `groupByKeyValue` of  `parseFunctionArguments`, which creates an empty group when seeing the comma. It seems like a simple fix would be to ignore the comma if it is trailing, this seems to work based on the tests I added.

### DIFF
--- a/src/rules/at-function-named-arguments/__tests__/index.js
+++ b/src/rules/at-function-named-arguments/__tests__/index.js
@@ -72,6 +72,16 @@ testRule(rule, {
     {
       code: `
       .b {
+        border: reset(
+          $value: 40px,
+        );
+      }
+      `,
+      description: "Always. Example: trailing comma after last argument."
+    },
+    {
+      code: `
+      .b {
         border: reset($value: 40px, $second-value: 10px, $color: 'black');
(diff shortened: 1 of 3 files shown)


kristerkari:
Thanks! The fix looks good.

kristerkari:
the fix is available in version 3.3.1 👍

thibaudcolas:
Thank you 👌 I can confirm it works on my project!

PR #272, opened 2018-09-23
Fix link markup in contribution guidelines

thibaudcolas:
The links were not rendering correctly because of the extra spaces.


kristerkari:
Thanks!

PR #279, opened 2018-10-23
Account for hyphens in function and mixin names

niksy:
Current rules won’t fix cases where function or mixin name uses hyphens, e.g.:

```scss
@function first-column-percentage-width($ratio) {
	@return percentage(nth($ratio, 1) / (nth($ratio, 2) - nth($ratio, 1)));
}
```


kristerkari:
Thanks! Could you also add tests for the hyphens in function and mixing names, that way it's easier to make sure that it won't break in the future. :)

kristerkari:
Awesome, looks good now

kristerkari:
the fix is available in v3.3.2

PR #280, opened 2018-10-25
Initial work on "at-mixin-argumentless-call-parentheses" fix

niksy:
This PR adds fix functionality to `at-mixin-argumentless-call-parentheses`. There is one problem, though: it should also depend on `at-mixin-parentheses-space-before` rule for spacing but currently there is no way to check against other rule. There is [`stylelint.utils.checkAgainstRule`](https://stylelint.io/developer-guide/plugins/#stylelintutilscheckagainstrule) helper utility, but it doesn’t know about other SCSS rules since [Jest Stylelint setup works only for one rule](https://github.com/kristerkari/stylelint-scss/blob/69a4c625009fde200c54556d11a3f18fa34cf29c/jest-setup.js#L23-L29). Does it make sense to upgrade Jest config to allow additional configuration to account for these changes?

### DIFF
--- a/src/rules/at-mixin-argumentless-call-parentheses/__tests__/index.js
+++ b/src/rules/at-mixin-argumentless-call-parentheses/__tests__/index.js
@@ -4,6 +4,7 @@ testRule(rule, {
   ruleName,
   config: ["never"],
   syntax: "scss",
+  fix: true,
 
   accept: [
     {
@@ -31,6 +32,9 @@ testRule(rule, {
       code: `
       @include foo () ;
     `,
+      fixed: `
+      @include foo ;
+    `,
       line: 2,
       message: messages.rejected("foo"),
       description: "With parens, no space between them."
@@ -38,6 +42,9 @@ testRule(rule, {
     {
       code: `
       @include foo ( ) ;
+    `,
+      fixed: `
+      @include foo ;
     `,
       line: 2,
       message: messages.rejected("foo"),
@@ -47,6 +54,9 @@ testRule(rule, {
       code: `
       @include foo( ) ;
     `,
+      fixed: `
+      @include foo ;
+    `,
       line: 2,
       message: messages.rejected("foo"),
       description: "With parens, no space before and space between them"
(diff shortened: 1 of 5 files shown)


kristerkari:
Thanks!



>  There is one problem, though: it should also depend on `at-mixin-parentheses-space-before` rule for spacing but currently there is no way to check against other rule.

I agree, the if the fix is applied, then it should not cause other rules to throw warnings.



> Does it make sense to upgrade Jest config to allow additional configuration to account for these changes?

Yes, the Jest config can be updated if needed for better testing.

niksy:
It looks like I was wrong in approach to this, Stylelint doesn’t support dependeant rules configuration (yet): https://github.com/stylelint/stylelint/issues/3210#issuecomment-372288465, https://github.com/stylelint/stylelint/issues/2532#issuecomment-302109499

Should we still create fixer for this and let users know caveats regarding using it?

kristerkari:
Yeah actually now I remember that there was already some discussion about it (the issue 3210). 

What jeddy3 said about rules not having to worry about introducing warnings for other rules is then the way we should do it.

src/rules/at-mixin-argumentless-call-parentheses/__tests__/index.js
```diff
@@ -30,6 +31,9 @@ testRule(rule, {
     {
       code: `
       @include foo () ;
+    `,
+      fixed: `
+      @include foo () ;
```
kristerkari: btw. the fixes here should not have the parentheses.

niksy:
@kristerkari changes pushed!

kristerkari:
Looks good now, thanks!

niksy:
@kristerkari should I mention that fix could clash with other rules?

kristerkari:
sure, I think that it's good to let the users know about it if they use the fix option.

niksy:
@kristerkari done!

kristerkari:
Awesome, this is going out in the next release. Thanks again!

Issue #197, opened 2018-01-29
Rule to enforce nesting of elements


hisuwh:
I want stylelint to complain if I do this:

```css
.something .something-else {
    width: 100px;
}
```

And be happy if I do this:
```css
.something {
    .something-else {
        width: 100px;
    }
}
```

Much like sass-lint's [force-element-nesting](https://github.com/sasstools/sass-lint/blob/master/docs/rules/force-element-nesting.md)

kristerkari:
Thanks! That sounds like a good rule idea

kristerkari:
This now available with ` selector-nest-combinators` rule in v3.5.0

I also opened a new issue to add more rules to enforce nesting:
https://github.com/kristerkari/stylelint-scss/issues/299

PR #298, opened 2019-01-09
[#197] - New rule to enforce nesting

lxsymington:
Added a new rule called `selector-nest-combinator` as well as a README and tests.

Currently the only options are `always` and `never`. This rule may be more useful with more options, such as `pseudo-only` or exceptions such as `attribute`.


kristerkari:
Thanks a lot @lxsymington! I will try to review this as soon as possible.

kristerkari:
Thanks again!

This is now available in v3.5.0

I also opened a new issue to add more rules to enforce nesting:
https://github.com/kristerkari/stylelint-scss/issues/299

lxsymington:
No problem! Thanks! I will try and have a look at that when I get some time.

Issue #300, opened 2019-01-15
selector-nest-combinators messaging


lxsymington:
#197 implemented selector-nest-combinators, however, the messages are malformed and therefore causing either unhandled errors or undefined output.

```
events.js:183
      throw er; // Unhandled 'error' event
      ^
TypeError: message.text.replace is not a function
    at orderedMessages.map.message (C:\git\vouchercloud-idl\IDL.Web.Vc\node_modules\stylelint\lib\formatters\stringFormatter.js:124:10)
    at Array.map (<anonymous>)
    at formatter (C:\git\vouchercloud-idl\IDL.Web.Vc\node_modules\stylelint\lib\formatters\stringFormatter.js:113:43)
    at results.reduce (C:\git\vouchercloud-idl\IDL.Web.Vc\node_modules\stylelint\lib\formatters\stringFormatter.js:182:15)
    at Array.reduce (<anonymous>)
    at module.exports (C:\git\vouchercloud-idl\IDL.Web.Vc\node_modules\stylelint\lib\formatters\stringFormatter.js:168:20)
    at module.exports (C:\git\vouchercloud-idl\IDL.Web.Vc\node_modules\stylelint\lib\formatters\verboseFormatter.js:8:16)
    at reporter (C:\git\vouchercloud-idl\IDL.Web.Vc\node_modules\gulp-stylelint\dist\reporter-factory.js:58:25)
    at C:\git\vouchercloud-idl\IDL.Web.Vc\node_modules\gulp-stylelint\dist\index.js:151:14
    at Array.map (<anonymous>)
    at passLintResultsThroughReporters (C:\git\vouchercloud-idl\IDL.Web.Vc\node_modules\gulp-stylelint\dist\index.js:150:44)
    at tryCallOne (C:\git\vouchercloud-idl\IDL.Web.Vc\node_modules\gulp-stylelint\node_modules\promise\lib\core.js:37:12)
    at C:\git\vouchercloud-idl\IDL.Web.Vc\node_modules\gulp-stylelint\node_modules\promise\lib\core.js:123:15
    at flush (C:\git\vouchercloud-idl\IDL.Web.Vc\node_modules\asap\raw.js:50:29)
    at _combinedTickCallback (internal/process/next_tick.js:132:7)
    at process._tickCallback (internal/process/next_tick.js:181:9)
```

PR #301, opened 2019-01-15
[#300] - Messaging fix

lxsymington:
The messaging for selector-nest-combinators was not functioning as
expected, this commit should fix those issues.


kristerkari:
Thanks!

PR #307, opened 2019-02-01
Handle interpolation in selector-nest-combinators

### LINKED ISSUE #306: false positive 'selector-nest-combinators'
Given a value 'always' and a stylesheet:

```
.class-name {
	#{if(&, "&", "")} {

	}	
}
```

The linter will report the error:

`'Expected combinator "}" of type "tag" to be in a nested form'`

kristerkari:
Fixes #306

### DIFF
--- a/src/rules/selector-nest-combinators/index.js
+++ b/src/rules/selector-nest-combinators/index.js
@@ -4,9 +4,10 @@ import { namespace, parseSelector } from "../../utils";
 export const ruleName = namespace("selector-nest-combinators");
 
 export const messages = utils.ruleMessages(ruleName, {
+  expectedInterpolation: `Expected interpolation to be in a nested form`,
   expected: (combinator, type) =>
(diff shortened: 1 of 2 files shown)


src/rules/selector-nest-combinators/index.js
```diff
@@ -20,21 +21,39 @@ export default function(expectation) {
...
+    const chainingTypes = [
+      "attribute",
+      "class",
+      "id",
+      "pseudo",
+      "tag",
+      "universal"
+    ];
+
+    const interpolationRe = /#{.+}$/;
```
lxsymington: Given that this is checking for the presence of any interpolation in the current selector, it's probably not particularly relevant, but it would generallly be better to use the non-greedy + operator? So that it would not match `#{$var} .selector li #{$var-2}`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#quantifiers

PR #318, opened 2019-04-16
Support stylelint v10

ricardogobbosouza:
- Removed: node.js 6.x support
- Update peer dependency stylelint v10.0.0

Resolve #316
Resolve #315

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "stylelint": "^10.0.1"
   },
   "peerDependencies": {
-    "stylelint": "^10.0.0"
+    "stylelint": "^8.0.0 || ^9.0.0 || ^10.0.0"
   },
   "engines": {
     "node": ">=8"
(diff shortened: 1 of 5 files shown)


kristerkari:
Thanks! I'll review

package.json
```diff
@@ -21,33 +21,33 @@
...
     "request": "^2.88.0",
     "rimraf": "^2.6.3",
-    "semver": "^5.6.0",
-    "stylelint": "^9.10.1"
+    "semver": "^6.0.0",
+    "stylelint": "^10.0.1"
   },
   "peerDependencies": {
-    "stylelint": "^8.0.0 || ^9.0.0"
+    "stylelint": "^10.0.0"
```
kristerkari: This should be `^8.0.0 || ^9.0.0 || ^10.0.0`. We are still supporting the old stylelint versions unless there is a breaking change somewhere in the `stylelint-scss` codebase.

ricardogobbosouza: @kristerkari You're right

kristerkari:
Thanks a lot! v3.6.0 is out with these changes

PR #324, opened 2019-05-24
Adding rule for unnecessary calls to quote()

rambleraptor:
The sass/linter project had a [suggestion for a linter rule](https://github.com/sass/linter/issues/37) that would call out unnecessary calls to quote().

This rules finds unnecessary calls to `quote()` and removes the call to `quote()` if autofixing is enabled.


rambleraptor:
@nex3 for your context

nex3:
It might also be a good idea to add a test case for passing a variable to `quote()`.

rambleraptor:
Great point! I'll do that + the checking to establish if a variable's contents are quoted / unquoted.

rambleraptor:
Added in new tests + functionality to handle variables.

I'm checking specifically the variable to see if the direct value is quoted or not. If the variable calls a function (which calls a function, which calls a function...), this rule will allow it (even if it ends up bringing a false-negative). I personally don't think that's improper behavior for the rule.

nex3:
I think that's fine. There are always going to be cases where we can't statically analyze the provenance of an argument.

src/rules/no-quoted-strings-inside-quote-function/README.md
```diff
@@ -0,0 +1,46 @@
+# no-quoted-strings-inside-quote-function
```
kristerkari: I think that the name should follow the same format as stylelint, e.g. `function-quote-no-quoted-strings-inside`:
https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules.md#function

src/rules/no-quoted-strings-inside-quote-function/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+Disallow quoted strings inside the quote function.
```
kristerkari: Should this link to the Sass docs for the quote function? People might not be familiar with it.

src/rules/no-quoted-strings-inside-quote-function/README.md
kristerkari: https://sass-lang.com/documentation/functions/string#quote

src/rules/no-quoted-strings-inside-quote-function/README.md
kristerkari: https://sass-lang.com/documentation/functions/string#quote

rambleraptor:
Comments addressed!

PR #325, opened 2019-05-24
Rule for using @each with maps properly

rambleraptor:
https://github.com/sass/linter/issues/33

sass/linter had a rule request for a linter that enforces proper loops over maps. This is a rule that checks for situations where users have:
- Done a loop using map-keys
- Grabbed the value for that key inside of the loop.


src/rules/each-key-value-single-line/README.md
```diff
@@ -0,0 +1,48 @@
+# each-key-value-single-line
```
kristerkari: Also this rule should follow the naming conventions from stylelint, e.g. `at-rule-each-*`:
https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules.md#at-rule-1

src/rules/each-key-value-single-line/README.md
```diff
@@ -0,0 +1,48 @@
...
+ Disallow situations where users loop over map-keys and fetch the value inside the loop.
+
+ ```scss
+ $font-weights: ("regular": 400, "medium": 500, "bold": 700);
+ @each $key in map-keys($font-weights) {
+   $value: map-get($font-weights, $key);
+/**        ↑
+ * This call should be consolidated into the @each call.
+ }
+  }
```
kristerkari: The syntax inside this SCSS block is broken, there is an extra `}` and the comment is not closed.

src/rules/each-key-value-single-line/README.md
```diff
@@ -0,0 +1,48 @@
...
+ The following patterns are *not* considered violations:
+
+ ```scss
+ $font-weights: ("regular": 400, "medium": 500, "bold": 700);
+ @each $key, $value in $font-weights {...}
+```
+
+ ```scss
+ $font-weights: ("regular": 400, "medium": 500, "bold": 700);
+ $other-weights: ("regular": 400, "medium": 500, "bold": 700);
```
kristerkari: Could you add unit test for this case?

src/rules/each-key-value-single-line/README.md
```diff
@@ -0,0 +1,48 @@
+# each-key-value-single-line
+
+ Disallow situations where users loop over map-keys and fetch the value inside the loop.
```
kristerkari: Is there some way yo make the description a bit clearer? Might be a bit confusing for some users.

kristerkari: Could be even this:

```
This is a rule that checks for situations where users have:

 - Done a loop using map-keys
 - Grabbed the value for that key inside of the loop.

```

rambleraptor:
Comments addressed!

kristerkari:
Thanks again! Both rules are available in v3.7.0

I had to rename this rule to `at-each-key-value-single-line`, so that it is consistent with the naming of the other rules in this plugin.

rambleraptor:
@nex3 should've pointed this out earlier, but this was merged in!

fuhlig:
Getting an error when setting this rule to `true`.

```
TypeError: Cannot read property 'replace' of undefined
    at orderedMessages.map.message (\node_modules\stylelint\lib\formatters\stringFormatter.js:124:10)
    at Array.map (<anonymous>)
    at formatter (\node_modules\stylelint\lib\formatters\stringFormatter.js:113:43)
    at results.reduce (\node_modules\stylelint\lib\formatters\stringFormatter.js:182:15)
    at Array.reduce (<anonymous>)
    at module.exports (\node_modules\stylelint\lib\formatters\stringFormatter.js:168:20)
    at prepareReturnValue (\node_modules\stylelint\lib\standalone.js:285:15)
    at globby.then.then.stylelintResults (\node_modules\stylelint\lib\standalone.js:274:14)
```

stylelint: 10.0.1
stylelint-scss: 3.7.0

kristerkari:
Thanks for reporting @fuhlig, I'll have a look.

kristerkari:
@fuhlig I released a fix in v3.8.0

fuhlig:
Looking good.
Thanks for the quick fix!

PR #326, opened 2019-05-31
Lint rule to check for unquote() receiving an unquoted string

rambleraptor:
The sass/linter repo has a [suggestion for a rule](https://github.com/sass/linter/issues/37) that would alert users using unquote on an already unquoted string.

This rules finds unnecessary calls to `unquote()` and removes the call to `unquote()` if autofixing is enabled.

(This is very much in line with #324)


rambleraptor:
@nex3 for your context

kristerkari:
Released in v3.8.0

PR #331, opened 2019-06-14
Ensure that all map keys are quoted

rambleraptor:
sass/linter has a [rule] that ensured that all map keys are quoted.

This replicates that rule.


rambleraptor:
@nex3

kristerkari:
Thanks! I was quite busy last weekend, but I'll review your PRs when I have a moment to look at them.

src/rules/map-keys-always-quoted/README.md
```diff
@@ -0,0 +1,26 @@
+# map-keys-always-quoted
```
kristerkari: Instead of `map-keys-always-quoted`, this rule could be called something like `map-keys-quotes` with options `always`/`never`:
https://github.com/stylelint/stylelint/blob/master/docs/user-guide/about-rules.md#about-rule-names

src/rules/map-keys-always-quoted/index.js
```diff
@@ -0,0 +1,83 @@
+import { utils } from "stylelint";
+import { namespace } from "../../utils";
+import valueParser from "postcss-value-parser";
+
+export const ruleName = namespace("map-keys-always-quoted");
+
+export const messages = utils.ruleMessages(ruleName, {
+  rejected: "Quote function used with an already-quoted string"
```
kristerkari: I forgot this when I was reviewing the previous rules, but the naming convention in stylelint error messages is `"Expected [something] [in some context]"` and `"Unexpected [something] [in some context]"`. We should try to stick to that message format in the future.

https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#determine-violation-messages

src/rules/map-keys-always-quoted/README.md
kristerkari: There is a similar rule in stylelint for `font-family` quotes:
https://github.com/stylelint/stylelint/blob/master/lib/rules/font-family-name-quotes/README.md

rambleraptor: I'm not sure if an always/never split would be necessary in this case.

It seems like you'd never want a situation where every key in your map is unquoted.

kristerkari: Right, in that case the only option could be `always`, like in the font-family rule.

rambleraptor: Is there a difference between using a "always" flag instead of a true/false flag?

kristerkari: IMO there are at least two things why it would make sense to use `always` instead of `true`.

1) Consistency. The other `-quotes` rules in stylelint use the `always`/`never` naming. (e.g. [font-family-name-quotes](https://github.com/stylelint/stylelint/blob/master/lib/rules/font-family-name-quotes/README.md) and [function-url-quotes](https://github.com/stylelint/stylelint/blob/master/lib/rules/function-url-quotes/README.md))
2) Ability to add more options in the future. If we use `always`, then we can add `always-*`, `never-*`, or `never` options if they are needed. Like in the `font-family-name-quotes` rule: https://github.com/stylelint/stylelint/blob/master/lib/rules/font-family-name-quotes/README.md#options

rambleraptor: That makes so much sense. Changes made!

kristerkari:
Thanks, released in 3.9.0:
https://github.com/kristerkari/stylelint-scss/releases/tag/3.9.0

PR #333, opened 2019-06-14
use scalar-color function instead of other color adjustment functions

rambleraptor:
sass/linter had a [rule](https://github.com/sass/linter/blob/master/lib/src/rules/use_scale_color.dart) that told users to use `scalar-color` instead of a variety of other functions.

This linter rule tells users to use the scalar-color function.


src/rules/function-scale-color-use-always/index.js
```diff
@@ -0,0 +1,52 @@
+import { utils } from "stylelint";
+import { namespace } from "../../utils";
+import valueParser from "postcss-value-parser";
+
+export const ruleName = namespace("function-scale-color-use-always");
+
+export const messages = utils.ruleMessages(ruleName, {
+  rejected: "Use the scale-color function."
```
kristerkari: As I pointed out in the other reviews, the naming convention in stylelint error messages is `"Expected [something] [in some context]"` and `"Unexpected [something] [in some context]"`. We should try to stick to that message format in the future.

https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#determine-violation-messages

src/rules/function-scale-color-use-always/README.md
```diff
@@ -0,0 +1,83 @@
+# function-scale-color-use-always
```
kristerkari: This name feels a bit strange, and is in a format that is not used in stylelint.

I'm not sure what would be a good name that would match stylelint's naming conventions.

rambleraptor: `scale-color` is meant to change a color by a relative amount, rather than a fixed amount.

What if the rule was called something like: `function-scale-color-relative`. It makes it clear it applies to the scale-color function and that the effect is to make sure it is used to make color changes relative?

kristerkari: Hmm... or maybe even more generic `function-color-relative` or `function-no-fixed-color`.

...to make it match a little bit the naming style of the rules in stylelint: `function-name-case`, `function-parentheses-newline-inside`, `property-no-vendor-prefix`, etc.

rambleraptor: I think `function-color-relative` would be a good name. Made the name change and sent it up.

kristerkari: Looks good now!

kristerkari:
Thanks, released in 3.9.0:
https://github.com/kristerkari/stylelint-scss/releases/tag/3.9.0

PR #342, opened 2019-07-09
selector-no-union-class-name: fix false positives for id/attr/pseudo

### LINKED ISSUE #341: `selector-no-union-class-name` rule false positive
The `selector-no-union-class-name` rule thinks there are errors here:

```scss
.icon {
  &:last-child { // Error on this line
    margin-left: 0.75rem;
  }

  &#divID { // Error on this line
    margin-left: 0.75rem;
  }
}
```

It seems the rule doesn't realize that pseudo-element selectors and ID selectors aren't class name unions.

cc @pipopotamasu

kristerkari:
Fixes #341

@nwoltman @pipopotamasu

### DIFF
--- a/src/rules/selector-no-union-class-name/__tests__/index.js
+++ b/src/rules/selector-no-union-class-name/__tests__/index.js
@@ -1,4 +1,4 @@
-import rule, { ruleName, messages } from "..";
+import rule, { messages, ruleName } from "..";
 
 testRule(rule, {
   ruleName,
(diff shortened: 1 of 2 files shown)


nwoltman:
This looks like it'll cover all the false positives in my codebase 👍

kristerkari:
Thanks for having a look, I think that this fix is good to go then!

mpg-yamato-murakami:
thx!

pipopotamasu:
thx!

PR #344, opened 2019-07-12
New rule: `at-if-no-null`

rambleraptor:
The [sass/linter](https://github.com/sass/linter/blob/master/lib/src/rules/use_falsey_null.dart) project had a rule to check for using `== null` or `!= null` inside @if statements.

This rule replicates that.

### DIFF
--- a/src/rules/at-if-no-null/README.md
+++ b/src/rules/at-if-no-null/README.md
@@ -4,7 +4,7 @@ Check for equality to null is unnecessarily explicit since `null` is falsey in S
 
 ```scss
 a {
-    @if ($x == null) {}
+    @if $x == null {}
 /**         ↑     ↑
  * == or != null is unncessary */
 }
@@ -12,20 +12,20 @@ a {
 
 ## Options
 
-always
+true
 
-### `always`
+### `true`
 
 The following patterns are considered warnings:
 ```scss
 a {
-    @if ($x == null) {}
+    @if $x == null {}
 }
 ```
 
 ```scss
 a {
-    @if ($x != null) {}
+    @if $x != null {}
 }
 ```
 
@@ -33,12 +33,12 @@ The following patterns are *not* considered warnings:
 
 ```scss
 a {
-    @if ($x) {}
+    @if $x {}
 }
 ```
 
 ```scss
 a {
-    @if not ($x) {}
+    @if not $x {}
 }
 ```
--- a/src/rules/at-if-no-null/README.md
+++ b/src/rules/at-if-no-null/README.md
@@ -1,1 +1,44 @@
 # at-if-no-null
+
+Check for equality to null is unnecessarily explicit since `null` is falsey in Sass.
+
+```scss
+a {
+    @if ($x == null) {}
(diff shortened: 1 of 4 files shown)


rambleraptor:
@nex3

src/rules/at-if-no-null/__tests__/index.js
```diff
@@ -0,0 +1,41 @@
...
+    }`,
+      description: "does not use the != null format"
+    },
+    {
+      code: `a {
+      @if not ($x) {}
+    }`,
+      description: "does not use the == null format"
+    }
+  ],
```
kristerkari: There could be some more tests with calculations and other conditions to make sure that there are no false positives/negatives.

src/rules/at-if-no-null/index.js
```diff
@@ -0,0 +1,44 @@
+import { namespace } from "../../utils";
+import { utils } from "stylelint";
+
+export const ruleName = namespace("at-if-no-null");
+
+export const messages = utils.ruleMessages(ruleName, {
+  equals_null: "Expected @if(statement) rather than @if(statement == null)",
+  not_equals_null: "Expected @if(!statement) rather than @if(statement != null)"
```
kristerkari: Are `@if not ($x)` and `@if(!$x)` both valid? Maybe the error messages and tests should have both examples?

src/rules/at-if-no-null/README.md
```diff
@@ -0,0 +1,44 @@
+# at-if-no-null
+
+Check for equality to null is unnecessarily explicit since `null` is falsey in Sass.
+
+```scss
+a {
+    @if ($x == null) {}
```
nex3: I'd like to discourage the use of parentheses with `@if` as well, so maybe these examples should omit them too?

src/rules/at-if-no-null/__tests__/index.js
rambleraptor: @nex3 any ideas on good tests or language constructs that users might be using in these if/else statements?

src/rules/at-if-no-null/index.js
rambleraptor: I changed the messages a little bit. How do they sound now?

src/rules/at-if-no-null/__tests__/index.js
nex3: When I'm looking for examples, I often look through big open-source corpuses like https://github.com/angular/components/tree/master/src/material or https://github.com/carbon-design-system/carbon/tree/master/packages.

src/rules/at-if-no-null/index.js
kristerkari: Sounds good now.

src/rules/at-if-no-null/__tests__/index.js
rambleraptor: Great idea! I didn't know of any large Sass codebases.

I grepped through carbon-design and came across a common pattern (if $x != null and expr). I made adjustments to the rule for that pattern.

PR #346, opened 2019-07-17
Fix selector-no-union-class-name throwing when parent.selector is undefined

### LINKED ISSUE #345: stylelint-scss selector-no-union-class-name breaks entire stylelint process
Reproduction css example:

```css

@mixin demo() {
   @content;
}

.a-selector {

  @include demo() { 
      button:hover & {
  	background:purple
     }
  }

}
```

If you run this code against the `selector-no-union-class-name` test it will crash the entire process.

I did some research and found out that the reason is that the parent of `button:hover & {` is `@include demo() { ` and therefore there is no `selector` property for the parent in this line:

https://github.com/kristerkari/stylelint-scss/blob/147b44cf10725fbf7b08a33885a6f7b7b8df4bb9/src/rules/selector-no-union-class-name/index.js#L38

undefinedSelector

The good news is that the `selector` is in the `parent` of the `parent`:
parent_parent

So probably we could just iterate over the parents until we find the closest `parent` with a selector` property.
If none can be found we just skip the test.

What do you think?

kristerkari:
Fixes #345

### DIFF
--- a/src/rules/selector-no-union-class-name/__tests__/index.js
+++ b/src/rules/selector-no-union-class-name/__tests__/index.js
@@ -95,6 +95,20 @@ testRule(rule, {
       }
       `,
       description: "ignores an ampersand chained with an attribute"
+    },
+    {
+      code: `
+      @mixin demo() {
+        @content;
+      }
+      .a-selector {
+        @include demo() {
+          button:hover & {
+  	        background:purple
+          }
+        }
+      }`,
+      description: "should not throw an error when using nesting (issue #345)"
     }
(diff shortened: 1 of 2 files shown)


jantimon:
Wow that was quick - maybe we can improv it just a little bit - i’ll send a small pull request on top of your change tomorrow

kristerkari:
> maybe we can improv it just a little bit - i’ll send a small pull request on top of your change tomorrow

Sure, just ping me when you have something ready. :+1:

PR #350, opened 2019-07-31
map-keys-quotes: don't require numbers to be quoted

### LINKED ISSUE #348: map-keys-quotes rule should allow ignoring numbers
Hi! `map-keys-quotes` is pretty useful, thanks for this addition. In my opinion, it should allow for numbers to stay unquoted though. For example in this color map:

```
$colors: (
  'primary': (
    250: #eef9ff,
    300: #bbe7ff,
    350: #b7d3e6,
    500: #1388db,
    750: #0f6bac,
    900: #2e4662,
  )
);
```

it doesn't make much sense to quote the keys of the 2nd level submap since they are numeric on purpose. However, it would be advantageous to still have string-like keys (such as `primary`) to be quoted. Would adding 'ignore' or similar option to this rule be feasible?

kristerkari:
Fixes #348

ping @certainlyakey @rambleraptor @nex3

### DIFF
--- a/src/rules/map-keys-quotes/__tests__/index.js
+++ b/src/rules/map-keys-quotes/__tests__/index.js
@@ -16,7 +16,7 @@ testRule(rule, {
       code: `
         $test: ('foo': 14px, 'bar': 25px);
       `,
-      description: "accepts strings without quotes"
+      description: "accepts strings with quotes"
     },
     {
       code: `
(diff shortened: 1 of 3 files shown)


src/rules/map-keys-quotes/__tests__/index.js
```diff
@@ -11,6 +11,40 @@ testRule(rule, {
         $test: ("foo": 14px, "bar": 25px);
       `,
       description: "accepts strings without quotes"
+    },
+    {
+      code: `
+        $test: ('foo': 14px, 'bar': 25px);
+      `,
+      description: "accepts strings without quotes"
```
nex3:
```suggestion
      description: "accepts strings with quotes"
```

?

kristerkari: Yeah I copy/pasted it from the other test :)

PR #351, opened 2019-08-05
New rule: at-conditional-rule-no-parentheses

rambleraptor:
sass/linter had a [rule request](https://github.com/sass/linter/issues/15) for a rule that would look for unnecessary parentheses in conditionals


src/rules/at-conditional-rule-no-parentheses/__tests__/index.js
```diff
@@ -0,0 +1,53 @@
...
+      code: `@if true {}
+      @else if true {}`,
+      description: "accepts @else if with no parentheses"
+    },
+    {
+      code: `@if true {}
+      @else if true {}
+      @else {}`,
+      description: "accepts @else unconditionally"
+    }
```
nex3: Maybe test that `@if (1 + 1) == 2` is allowed?

src/rules/at-conditional-rule-no-parentheses/index.js
```diff
@@ -0,0 +1,69 @@
+import { utils } from "stylelint";
+import { namespace } from "../../utils";
+
+export const ruleName = namespace("at-conditional-rule-no-parentheses");
+
+export const messages = utils.ruleMessages(ruleName, {
+  rejected: "Do not use () to surround statements for @-rules"
```
kristerkari: This should follow the stylelint naming, so either `Expected ..` or `Unexpected...`

src/rules/at-conditional-rule-no-parentheses/index.js
```diff
@@ -0,0 +1,69 @@
...
+  });
+}
+
+function fix(atrule) {
+  const regex = /(if)? ?\((.*)\)/;
+
+  // 2 regex groups: 'if ' and cond.
+  const groups = atrule.params.match(regex).slice(1);
+
+  atrule.params = Array.from(new Set(groups)).join(" ");
```
kristerkari: I wonder if `Array.from` and `Set` are supported in the environments that stylelint supports?

src/rules/at-conditional-rule-no-parentheses/__tests__/index.js
rambleraptor: Great idea! Added + made a fix.

rambleraptor:
@kristerkari ptal!

src/rules/at-conditional-rule-no-parentheses/index.js
rambleraptor: I'm not entirely sure. It looks like stylelint-scss has lodash bundled, so it's probably a better idea to use that to avoid compatibility issues.

src/rules/at-conditional-rule-no-parentheses/README.md
```diff
@@ -0,0 +1,43 @@
+# at-conditional-rule-no-parentheses
```
kristerkari: Maybe the name could be changed to `at-rule-conditional-no-parentheses`, so that it can be put under the `at-rule` category that many stylelint rules are using. What do you think @rambleraptor ?

src/rules/at-conditional-rule-no-parentheses/index.js
```diff
@@ -0,0 +1,70 @@
...
+      actual: primary
+    });
+
+    if (!validOptions) {
+      return;
+    }
+
+    root.walkAtRules(atrule => {
+      // Check if this is a conditional rule.
+      if (!conditional_rules.includes(atrule.name)) {
```
kristerkari: This could be changed to use `_.includes`, just in case `includes` would not be supported somewhere.

PR #357, opened 2019-08-12
New rule: No non numeric dimensions

rambleraptor:
The old sass/linter had a [rule](https://github.com/sass/linter/blob/master/test/rules/non_numeric_dimension_test.dart) to ensure that string interpolated values didn't have units attached to them (because they'd be strings instead of numbers).

This rule replicates that. The tests are the same for the sass/linter rule + this rule, so I have good confidence it'll give the same user experience.


rambleraptor:
@nex3 fyi

src/rules/no-non-numeric-dimensions/__tests__/index.js
```diff
@@ -0,0 +1,54 @@
+import rule, { ruleName, messages, units } from "..";
+
+testRule(rule, {
+  ruleName,
+  config: [true],
+  syntax: "scss",
+  accept: loopOverUnits({
+    code: `
+    p {
+      padding: "1" * 1%unit%;
```
nex3:
```suggestion
      padding: 1 * 1%unit%;
```

...otherwise you're multiplying a string by a number which won't work.

src/rules/no-non-numeric-dimensions/index.js
```diff
@@ -0,0 +1,127 @@
...
+  "dppx",
+  "x",
+
+  // Flexible lengths:
+  // https://www.w3.org/TR/css-grid-1/#fr-unit
+  "fr",
+
+  // Percentage:
+  // https://www.w3.org/TR/css-values-4/#percentages
+  "%"
```
nex3: `#{...}%` isn't valid syntax anyway, so you may as well exclude this.

src/rules/no-non-numeric-dimensions/index.js
```diff
@@ -0,0 +1,127 @@
+import { utils } from "stylelint";
+import { namespace } from "../../utils";
+import valueParser from "postcss-value-parser";
+
+export const ruleName = namespace("no-non-numeric-dimensions");
+
+export const messages = utils.ruleMessages(ruleName, {
+  rejected: unit => `Expected "$value * 1${unit}" instead of "#{value}${unit}"`
```
nex3: Consider also suggesting that they write `$value` in terms of `unit` to begin with. While `$value * 1${unit}` will mostly work, it's likely that their overall logic will be more coherent if they start out using the correct units.

nex3:
It's probably also a good idea to test that this only applies to unquoted strings.

rambleraptor:
good call on the test! added a test case for that.

src/rules/no-non-numeric-dimensions/__tests__/index.js
```diff
@@ -19,6 +19,11 @@ testRule(rule, {
     {
       code: "$pad: 2; $doublePad: #{$pad}pxx;",
       description: "does not report lint when no understood units are used"
+    },
+    {
+      code: `$pad: "2";
+      $string: "#{$pad}px";`,
+      description: "does not report lint when no understood units are used"
```
nex3: Maybe "when the string is quoted"?

rambleraptor: image

rambleraptor:
@kristerkari - ptal when you can!

kristerkari:
> @kristerkari - ptal when you can!

Sorry, I have been really busy for the last 2 weeks. 

I think that the new rules are okay, but only thing that remains is to think about the naming, are the rules names matching the stylelint naming conventions? Are the rules in the right "category", etc. 

I will try to think about the names this weekend, and give suggestions for the names if I can come up with any.

src/rules/index.js
```diff
@@ -77,6 +78,7 @@ export default {
   "media-feature-value-dollar-variable": mediaFeatureValueDollarVariable,
   "no-dollar-variables": noDollarVariables,
   "no-duplicate-dollar-variables": noDuplicateDollarVariables,
+  "no-non-numeric-dimensions": noNonNumericDimensions,
```
kristerkari: I have been thinking a bit about the name. It should probably start with the "category", e.g. [number-max-precision](https://github.com/stylelint/stylelint/blob/master/lib/rules/number-max-precision/README.md), [length-zero-no-unit](https://github.com/stylelint/stylelint/blob/master/lib/rules/length-zero-no-unit/README.md) as examples from stylelint.

How about `dimension-no-non-numeric-values` or something similar @rambleraptor ?

src/rules/no-non-numeric-dimensions/README.md
```diff
@@ -0,0 +1,93 @@
+# at-non-numeric-dimensions
```
kristerkari: Need to remember to update the title too...

src/rules/index.js
kristerkari: Example of similar naming from stylelint: [shorthand-property-no-redundant-values](https://github.com/stylelint/stylelint/blob/master/lib/rules/shorthand-property-no-redundant-values/README.md)

PR #365, opened 2019-09-05
Fix error reporting for dimension-no-non-numeric-values

### LINKED ISSUE #363: `dimension-no-non-numeric-values` throws error
Using the [demension-no-non-numeric-values](https://github.com/kristerkari/stylelint-scss/tree/master/src/rules/dimension-no-non-numeric-values) introduced in 3.10 I get the following error

```
// config
'scss/dimension-no-non-numeric-values': true

// stylesheet
$value: 4;

p {
  padding: #{value}px;
}
```

```
TypeError: message.text.replace is not a function
    at orderedMessages.map.message (\node_modules\stylelint\lib\formatters\stringFormatter.js:124:10)
    at Array.map (<anonymous>)
    at formatter (\node_modules\stylelint\lib\formatters\stringFormatter.js:113:43)
    at results.reduce (\node_modules\stylelint\lib\formatters\stringFormatter.js:182:15)
    at Array.reduce (<anonymous>)
    at module.exports (\node_modules\stylelint\lib\formatters\stringFormatter.js:168:20)
    at prepareReturnValue (\node_modules\stylelint\lib\standalone.js:285:15)
    at globby.then.then.stylelintResults (\node_modules\stylelint\lib\standalone.js:274:14)
```

The code inside the stylesheet does not matter. Setting the rule to `true` causes the error.

kristerkari:
Fix following issues related to the `dimension-no-non-numeric-values` rule:

- Rejected message was throwing an error because it was called without the unit.
- Change highlighted error to point to the unit that is used with the interpolation, instead of it pointing to the property name. That makes much more sense if you are using this rule with CSS shorthand properties.

Fixes #363

Please review @rambleraptor @fuhlig

### DIFF
--- a/src/rules/dimension-no-non-numeric-values/__tests__/index.js
+++ b/src/rules/dimension-no-non-numeric-values/__tests__/index.js
@@ -1,4 +1,4 @@
-import rule, { ruleName, messages, units } from "..";
+import rule, { messages, ruleName, units } from "..";
 
 testRule(rule, {
   ruleName,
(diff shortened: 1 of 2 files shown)


kristerkari:
Skärmavbild 2019-09-05 kl  21 16 11

PR #369, opened 2019-09-16
Added: support for stylelint version 11

### LINKED ISSUE #368: support stylelint@11
:star: https://github.com/kristerkari/stylelint-scss/blob/master/package.json#L47

bjankord:
Resolves #368

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,10 +41,10 @@
     "request": "^2.88.0",
     "rimraf": "^3.0.0",
     "semver": "^6.3.0",
-    "stylelint": "^10.1.0"
+    "stylelint": ">=10.1.0"
   },
   "peerDependencies": {
-    "stylelint": "^8.0.0 || ^9.0.0 || ^10.0.0"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -41,10 +41,10 @@
     "request": "^2.88.0",
     "rimraf": "^3.0.0",
     "semver": "^6.3.0",
-    "stylelint": "^10.1.0"
+    "stylelint": ">=10.1.0"
```
bjankord: I went with this syntax but could see updating this to `^11.0.0`.

kristerkari: Not a big thing, but would be better to use `^11.0.0` for consistency.

kristerkari:
Thanks! Test seems to be passing, so this looks good. I'll release a new version tomorrow.

package.json
bjankord: Updated in this commit: https://github.com/kristerkari/stylelint-scss/pull/369/commits/830a85a659c59ac27969e8fab6b6bca8a13cac54

PR #381, opened 2019-10-16
Add rule allowing @use and @font-face

Benjamin-Schodts:
Add support for @use (https://sass-lang.com/documentation/at-rules/use) and @font-face (https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face)

### DIFF
--- a/src/rules/at-rule-no-unknown/index.js
+++ b/src/rules/at-rule-no-unknown/index.js
@@ -11,6 +11,7 @@ const sassAtRules = [
   "else if",
   "error",
   "extend",
+  "font-face",
   "for",
   "function",
   "if",
@@ -19,6 +20,7 @@ const sassAtRules = [
   "media",
   "mixin",
   "return",
+  "use",
   "warn",
   "while"
 ];


src/rules/at-rule-no-unknown/index.js
```diff
@@ -11,6 +11,7 @@ const sassAtRules = [
   "else if",
   "error",
   "extend",
+  "font-face",
```
tysongach: `@font-face` is a CSS at-rule, not Sass, and [stylelint already accounts for it](https://github.com/stylelint/stylelint/blob/19f987e5bb3bf62d8ce634bd8dfe8482d5d76d4a/lib/reference/keywordSets.js#L537).

kristerkari: Yeah, I don't think that it should be part of this PR. :)

kristerkari:
Thanks @Benjamin-Schodts and @tysongach for this, I've merged https://github.com/kristerkari/stylelint-scss/pull/385 instead, but please let me know if there is something else that is needed.

PR #382, opened 2019-10-17
New rule: use @forward before @use

rambleraptor:
The [Sass docs](https://sass-lang.com/documentation/at-rules/forward) recommend using @forward before @use. This linter rule enforces that.


rambleraptor:
@nex3

src/rules/at-forward-before-use/index.js
```diff
@@ -0,0 +1,39 @@
+import { utils } from "stylelint";
+import { namespace } from "../../utils";
+
+export const ruleName = namespace("at-forward-before-use");
+
+export const messages = utils.ruleMessages(ruleName, {
+  rejected: "Forward function should be used before use function"
```
kristerkari: Are these called functions, at-rules, or something else in this context, what's the correct term for `@use` and `@forward`?

kristerkari: Also, this should follow the stylelint message guideline:

messages.expected = `Expected something to be something`

or

messages.rejected = `Unexpected something`

kristerkari:
A couple of small comments, but other than that looks good! :+1:

rambleraptor:
Comments addressed!

src/rules/at-forward-before-use/__tests__/index.js
```diff
@@ -0,0 +1,31 @@
...
+        @use "src/list"
+        @forward "src/list"
+      `,
+      description:
+        "does not accept use statements that are before forward statements",
+      message: messages.rejected,
+      line: 2
+    }
+  ]
+});
```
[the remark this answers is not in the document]
rambleraptor: Great idea. Addressed!

src/rules/at-forward-before-use/index.js
```diff
@@ -0,0 +1,40 @@
...
+    if (!validOptions) {
+      return;
+    }
+
+    const useRules = [];
+
+    root.walkAtRules(/forward|use/, decl => {
+      if (
+        decl.name === "forward" &&
+        useRules.includes(decl.params.replace(/"/g, ""))
```
nex3: Won't this fail for rules with clauses (`show`/`hide`/`as`/`with`), or rules that use `'` as their quote characters?

Does PostCSS provide a way to run an expression-level parser over the params, so you can consume a string according to precise CSS syntax?

kristerkari: @rambleraptor please add more unit tests if there are doubts that the code path does not work in all cases.

kristerkari: ping @rambleraptor

stof:
should this be autofixable (at least when there are no other rules between them) ?

nex3:
FYI @rambleraptor is no longer doing 20% work with the Sass team, so if someone else could take this over and land it that would be great.

PR #385, opened 2019-10-22
Enable two new Sass at-rules: @forward and @use.

srawlins:
This can probably replace #381.

See announcement of `@forward` and `@use` here: http://sass.logdown.com/posts/7858341-the-module-system-is-launched


kristerkari:
Thanks!

PR #395, opened 2019-11-08
at-rule-no-unknown: remove duplicate looping logic

vseventer:
This PR fixes a bug in `at-rule-no-unknown`. `utils.checkAgainstRule` and `root.walkAtRules` are both called, and both loop over rules. This leads to `n^2` warnings rather than `n`.

For example, linting the file with two unknowns:
```css
@foo { }
@bar { }
```
Leads to four warnings running:
```txt
1:1  Unexpected unknown at-rule "@foo" (at-rule-no-unknown)
1:1  Unexpected unknown at-rule "@foo" (at-rule-no-unknown)
2:1  Unexpected unknown at-rule "@bar" (at-rule-no-unknown)  
2:1  Unexpected unknown at-rule "@bar" (at-rule-no-unknown)
```

I wasn't sure how to add tests for this since it never failed any to begin with. If updated tests are required, I would appreciate any pointers!

### DIFF
--- a/src/rules/at-rule-no-unknown/__tests__/index.js
+++ b/src/rules/at-rule-no-unknown/__tests__/index.js
@@ -1,4 +1,9 @@
+import postcss from "postcss";
 import rule, { ruleName, messages } from "..";
+
+function logError(err) {
+  console.log(err.stack); // eslint-disable-line no-console
+}
 
 testRule(rule, {
   ruleName,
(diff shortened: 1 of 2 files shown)


kristerkari:
Thanks @vseventer! Looks good, but could you also add a unit test that would call the rule with the example input and then check how many warnings there are? 

e.g.
https://github.com/kristerkari/stylelint-scss/blob/0c77143435787dab569353ea5530def79638e232/src/rules/partial-no-import/__tests__/index.js#L12-L23

vseventer:
Thanks for the pointer @kristerkari - added a test.

kristerkari:
Thanks!

PR #405, opened 2019-11-17
Remove the unused request and semver devDependencies.

XhmikosR:


XhmikosR:
BTW it seems github-contributors-list isn't used either? I mean, if you use it globally, it shouldn't be listed in devDependencies. Otherwise there should be an npm script :)

kristerkari:
>  Otherwise there should be an npm script

There should be a script for it :)

XhmikosR:
Please add it when you can :)

PR #406, opened 2019-11-17
Reduce lodash usage.

### LINKED ISSUE #399: Reduce lodash usage
I have a working patch I'll submit it later

XhmikosR:
use the individual packages and replace the following methods with their native counterparts:

* assign
* head
* includes
* isUndefined
* repeat
* uniq

Fixes #399

### DIFF
--- a/src/utils/parseFunctionArguments.js
+++ b/src/utils/parseFunctionArguments.js
@@ -1,4 +1,3 @@
-import { head } from "lodash";
 import valueParser from "postcss-value-parser";
 
 export function groupByKeyValue(nodes) {
@@ -56,7 +55,7 @@ export function mapToKeyValue(nodes) {
     return acc;
   }, []);
 
-  return head(keyVal);
+  return keyVal[0];
 }
 
 export function parseFunctionArguments(value) {
@@ -66,7 +65,7 @@ export function parseFunctionArguments(value) {
     return [];
   }
 
-  return head(
-    parsed.nodes.map(node => groupByKeyValue(node.nodes).map(mapToKeyValue))
-  );
+  return parsed.nodes.map(node =>
+    groupByKeyValue(node.nodes).map(mapToKeyValue)
+  )[0];
 }
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -1,6 +1,6 @@
 "use strict"; // eslint-disable-line strict
 
-const _ = require("lodash");
+const get = require("lodash.get");
 const stylelint = require("stylelint");
 
 global.testRule = (rule, schema) => {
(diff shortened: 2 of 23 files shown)


src/utils/parseFunctionArguments.js
```diff
@@ -56,7 +55,7 @@ export function mapToKeyValue(nodes) {
     return acc;
   }, []);
 
-  return head(keyVal);
+  return keyVal[0];
```
XhmikosR: Need a confirmation if this can be null or undefined. If so I should add a fallback `[]`

src/utils/parseFunctionArguments.js
```diff
@@ -66,7 +65,7 @@ export function parseFunctionArguments(value) {
     return [];
   }
 
-  return head(
-    parsed.nodes.map(node => groupByKeyValue(node.nodes).map(mapToKeyValue))
-  );
+  return parsed.nodes.map(node =>
+    groupByKeyValue(node.nodes).map(mapToKeyValue)
+  )[0];
```
XhmikosR: Same here

stof:
the `lodash.*` packages are unmaintained (as can be seen by their version numbers, which are far behind). Why switching to them rather than importing the sub-packages from lodash (i.e. `lodash/get`), which are the recommended way ?

XhmikosR:
https://packagephobia.now.sh/result?p=lodash

XhmikosR:
Also they are not unmaintained.

stof:
@XhmikosR The size is why I suggested importing lodash modules, not the main index.

Regarding maintenance, `lodash.get` was last published 3 years ago for instance. Same is true for other packages as well (the only lodash-modularized packages released this year are the one affected by a security issue, and only that security patch was applied, not all other improvements AFAIK).
and the lodash roadmap says that they will be dropped entirely for lodash 5: https://github.com/lodash/lodash/wiki/Roadmap
the maintainer also said multiple times in comments that he consider them deprecated. see https://github.com/lodash/lodash/issues/3943#issuecomment-418998293 for instance.

XhmikosR:
Well, this is lodash 4.x. And if the time comes, we can just replace those with our utils.

Importing a huge package is never better.

stof:
https://github.com/lodash/lodash.com/pull/214 is adding a new page on the website about the fact that they are deprecated. It is explaining why importing multiple per-method packages usually end up with a bigger bundle size than importing multiple modules from the `lodash` package (short story: internal lodash helpers are duplicated in each per-method package while importing from the `lodash` packages will import the same helpers).
When importing modular lodash, you *install* a big package, but you don't *import* a big package.

XhmikosR:
This is not a web app. We are not bundling anything.

PR #408, opened 2019-11-17
Fix coveralls npm script

XhmikosR:
EDIT: This won't be needed if we merge #413.


XhmikosR:
This won't be needed if we merge #413.

PR #410, opened 2019-11-17
sassValueParser/index.js: fix duplicate character in regex

XhmikosR:
`/` is already present in the character list.

Found with https://lgtm.com/projects/g/kristerkari/stylelint-scss/alerts?mode=list


XhmikosR:
@kristerkari Personally I'd add lgtm to the repo; it's owned by GitHub now. I'd just disable PR comments :)

PR #414, opened 2019-11-19
Remove babel.

### LINKED ISSUE #400: Remove babel
I can't see any reason to keep using babel if you target modern Node.js versions and should simplify things a lot. Also, jest should work out of the box IIRC with import/exports.

Feel free to ping me if anyone makes a WIP branch :)

XhmikosR:
This should be looked into after the other PRs are merged because it will require a major version bump.

But while at it, we should require Node.js >= 8.10.0 for Object spread.

Do note that I had to use `postcss-media-query-parser`'s `default` object in the require.

Fixes #400


XhmikosR:
@kristerkari any thoughts? :)

kristerkari:
@XhmikosR looks good, but I'm still hesitating a bit about when to do the breaking change. :)

XhmikosR:
I guess we'll need to wait until the next major version of stylelint. When that time comes, we might be able to use native modules.

XhmikosR:
I'm gonna close this and just keep the branch.

PR #416, opened 2019-11-19
ES6-ify

XhmikosR:
We should probably enforce these later via ESLint.


XhmikosR:
@kristerkari we could probably merge this for this minor version, it should be OK, but it's your call.

XhmikosR:
#418 and #421 should be safe too. Node.js 8.0.0 ships with npm 5.0.0 which supports the change in #418.

Issue #397, opened 2019-11-17
Add a lint for using global function names


nex3:
Now that the module system has launched, functions that are available from modules should be called using those modules rather than using their global aliases. The canonical list of which functions have moved to which modules is [in the module system proposal](https://github.com/sass/sass/blob/master/accepted/module-system.md#built-in-modules-1).

Maybe the lint should also link to the [Sass migrator documentation](https://sass-lang.com/documentation/cli/migrator) to help users migrate automatically?

cc @rambleraptor

kristerkari:
New rule released in v3.16.0:
https://github.com/kristerkari/stylelint-scss/releases/tag/3.16.0

PR #425, opened 2020-01-03
Linter for global function names

rambleraptor:
Issue: https://github.com/kristerkari/stylelint-scss/issues/397

This creates a linter for global function names.

Couple things:

1) I didn't know how comprehensive to make the docs + test suite. Putting in examples for every rule seemed excessively verbose.

2) I purposefully didn't put in auto-fix functionality, since we're encouraging people to use the migrator. Does that seem safe?


rambleraptor:
@nex3

src/rules/no-global-function-names/__tests__/index.js
```diff
@@ -0,0 +1,51 @@
...
+
+testRule(rule, {
+  ruleName,
+  config: [true],
+  syntax: "scss",
+
+  accept: [
+    {
+      code: `
+      @use color;
```
nex3:
```suggestion
      @use "sass:color";
```

Also below.

src/rules/no-global-function-names/__tests__/index.js
```diff
@@ -0,0 +1,51 @@
...
+      a {
+       background: color.red(#6b717f);
+      }
+    `,
+      description: "Non-global function with different name"
+    }
+  ],
```
nex3: I'd also add accept tests for using a global function that's still allowed (such as `rgb()` and/or `if()`).

rambleraptor:
changes made!

kristerkari:
Looks good!

kristerkari:
Hmm.. maybe we should generate test cases for each function? If we don't test them then the test coverage will suffer as there will be so many lines that are not tested.

rambleraptor:
I had some spare time and put in a whole bunch more test cases. There's more that need to be put in, but this is a lot more than we had earlier.

kristerkari:
I think that this is good now. I need to look through the changes once more before merging.

PR #428, opened 2020-01-12
Update stylelint to v13.0.0

### LINKED ISSUE #433: Stylelint v13 support
Can you please add support for stylelint v13

XhmikosR:
Requires #427 merged and tests to be fixed

Closes #433

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,10 +40,10 @@
     "postcss-scss": "^2.0.0",
     "prettier": "^1.19.1",
     "rimraf": "^3.0.0",
-    "stylelint": "^12.0.0"
+    "stylelint": "^13.0.0"
   },
   "peerDependencies": {
-    "stylelint": "^8.0.0 || ^9.0.0 || ^10.0.0 || ^11.0.0 || ^12.0.0"
+    "stylelint": "^8.0.0 || ^9.0.0 || ^10.0.0 || ^11.0.0 || ^12.0.0 || ^13.0.0"
   },
   "engines": {
     "node": ">=8"
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "url": "https://github.com/kristerkari/stylelint-scss/issues"
   },
   "dependencies": {
+    "lodash.get": "^4.4.2",
     "lodash.isboolean": "^3.0.3",
     "lodash.isregexp": "^4.0.1",
     "lodash.isstring": "^4.0.1",
@@ -34,7 +35,6 @@
     "jest": "^25.1.0",
     "jest-cli": "^25.1.0",
     "lint-staged": "^10.0.2",
-    "lodash.get": "^4.4.2",
     "npmpub": "^5.0.0",
     "postcss": "^7.0.26",
     "postcss-scss": "^2.0.0",
--- a/README.md
+++ b/README.md
@@ -20,8 +20,6 @@ The plugin follows stylelint's guidelines (about [rule names](https://stylelint.
 
 stylelint-scss is a plugin for [stylelint](https://stylelint.io/user-guide), so it's meant to be used with it.
 
-**Node.js v6 or newer** is required. That's because stylelint itself [doesn't support Node.js versions below 6](https://github.com/stylelint/stylelint/issues/2996#issuecomment-352663375).
-
 First, install stylelint-scss (and stylelint, if you haven't done so yet) via NPM:
 
 ```
--- a/package-lock.json
+++ b/package-lock.json
@@ -1546,21 +1546,31 @@
         }
       }
     },
-    "@mrmlnc/readdir-enhanced": {
-      "version": "2.2.1",
-      "resolved": "https://registry.npmjs.org/@mrmlnc/readdir-enhanced/-/readdir-enhanced-2.2.1.tgz",
-      "integrity": "sha512-bPHp6Ji8b41szTOcaP63VlnbbO5Ny6dwAATtY6JTjh5N2OLrb5Qk/Th5cRkRQhkWCt+EJsYrNB0MiL+Gpn6e3g==",
+    "@nodelib/fs.scandir": {
+      "version": "2.1.3",
+      "resolved": "https://registry.npmjs.org/@nodelib/fs.scandir/-/fs.scandir-2.1.3.tgz",
+      "integrity": "sha512-eGmwYQn3gxo4r7jdQnkrrN6bY478C3P+a/y72IJukF8LjB6ZHeB3c+Ehacj3sYeSmUXGlnA67/PmbM9CVwL7Dw==",
       "dev": true,
       "requires": {
-        "call-me-maybe": "^1.0.1",
-        "glob-to-regexp": "^0.3.0"
+        "@nodelib/fs.stat": "2.0.3",
+        "run-parallel": "^1.1.9"
       }
     },
     "@nodelib/fs.stat": {
-      "version": "1.1.3",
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -41,13 +41,13 @@
     "postcss-scss": "^2.0.0",
     "prettier": "^1.19.1",
     "rimraf": "^3.0.0",
-    "stylelint": "^12.0.0"
+    "stylelint": "^13.0.0"
   },
   "peerDependencies": {
-    "stylelint": "^8.0.0 || ^9.0.0 || ^10.0.0 || ^11.0.0 || ^12.0.0"
+    "stylelint": "^8.0.0 || ^9.0.0 || ^10.0.0 || ^11.0.0 || ^12.0.0 || ^13.0.0"
```
XhmikosR: BTW this is getting too long. We should probably drop some versions from here.

XhmikosR:
@hudochenkov: maybe you could have a quick look? It seems all of the percent-placeholder-pattern tests fail with v13.0.0. After some quick debugging it seems the cause it's https://github.com/stylelint/stylelint/pull/4483

hudochenkov:
Your discovery is correct. This plugin uses internals of stylelint, which are not a public API.

https://github.com/kristerkari/stylelint-scss/blob/26097c6fd72aab40d76a90f176097ccd57a6c347/src/rules/percent-placeholder-pattern/index.js#L44-L52

https://github.com/kristerkari/stylelint-scss/blob/26097c6fd72aab40d76a90f176097ccd57a6c347/src/utils/isStandardSelector.js

https://github.com/kristerkari/stylelint-scss/blob/26097c6fd72aab40d76a90f176097ccd57a6c347/src/utils/isStandardRule.js

I would suggest copy required utils from stylelint (`isStandardSelector`, `isStandardRule`) prior change we made in https://github.com/stylelint/stylelint/pull/4483.

stylelint concerns about standard CSS only, that's why we have these utils. `stylelint-scss` expand concern by allowing Sass syntax. That's why utils should be different for both projects.

XhmikosR:
Thanks for the reply :) TBH I think stylelint-scss should not duplicate everything, hence why I made #430, because I see a lot seemingly identical code.

That being said, for the time being, I will try first your suggestion so that we can update to the latest version and leave the other refactoring changes for later :)

hudochenkov:
> TBH I think stylelint-scss should not duplicate everything, hence why I made #430, because I see a lot seemingly identical code.

This comes with a price, that something could break even with patch release of stylelint :)

I think you need to change `isStandardRule` only for now.

XhmikosR:
I was under the impression that utils were part of the public API too. Is there a list of things that are officially part of the public API?

hudochenkov:
https://stylelint.io/developer-guide/plugins#stylelintutils

XhmikosR:
The thing is that `stylelint.utils` is pretty vague IMHO. I'd need to print the object to see this. Would be nice if the docs covered this somehow.

XhmikosR:
Or does it mean that only the mentioned utils methods are exposed?

hudochenkov:
I think only exposed methods are available.

XhmikosR:
Hmm, then I guess we need the opposite of #430 so that we are safe. Have you guys thought about exposing the other utils too? Currently there's like 500 lines of code and 15 files that are duplicated, and if we add more files this will keep increasing.

hudochenkov:
There were never discussion about this. You can start one in stylelint repo :)

XhmikosR:
@kristerkari I think this is ready for review after #427 is merged. I added the upstream code and tests before the breaking change.

kristerkari:
Thanks! I've been away from home for the weekend and at work today the whole day, but I'll try to get this reviewed asap.

kristerkari:
btw. I merged #427

XhmikosR:
Rebased. I'm not sure if the new upstream files I backported make all sense; I see code for Less etc.

XhmikosR:
I split this for easier review.

XhmikosR:
@kristerkari frinedly ping for this and the other open PRs

package.json
```diff
@@ -34,19 +35,18 @@
...
     "npmpub": "^5.0.0",
     "postcss": "^7.0.26",
     "postcss-scss": "^2.0.0",
     "prettier": "^1.19.1",
     "rimraf": "^3.0.0",
-    "stylelint": "^12.0.0"
+    "stylelint": "^13.0.0"
   },
   "peerDependencies": {
-    "stylelint": "^8.0.0 || ^9.0.0 || ^10.0.0 || ^11.0.0 || ^12.0.0"
```
kristerkari: Hmm... I'm still not sure if it's a good idea to drop support for everything else than the newest version.

XhmikosR: Well, the thing sis that

1. what was the last time you tested stylelist-scss master with stylelint 8.0.0? Or any version other than the latest?
2. if you are going to make sure every future version works with stylelint, why don't you just use `>=`?
3. This should be a major release IMO and we should try to merge #429 too, which makes it  a good time to remove this

XhmikosR: I can revert it BTW, my initial patch didn't change this.

pustovalov:
WDYT about drop node 8 support?
https://github.com/stylelint/stylelint/pull/4500

XhmikosR:
The thing is that it seems stylelint-scss works with Node.js 8.x and maybe old stylelint versions.

IMHO we should grab this opportunity and:

1. drop Node.js < 10 support
2. drop babel (see #429)
3. reduce the supported stylelint versions in peerDependencies since we only test the latest

XhmikosR:
Friendly ping @kristerkari ^^

package.json
kristerkari: > what was the last time you tested stylelist-scss master with stylelint 8.0.0? Or any version other than the latest?

I don't think that you really understand my point of view here. I don't really care about stylelint version 8, and I don't know if it has even any users other than some outdated projects. I just don't think that it is a good idea to drop the support for 5 versions at the same time, as the chances that there are people using some those versions is really high.

I actually have one React Native project where I had to lock the stylelint version to v9.x because newer stylelint versions depend on some dependency that uses `@types/node` which messes up the Typescript typings for React Native in my project and causes Typescript to throw errors.

I've ran into similar issues with other npm packages, where I could not update to the latest version of a package because it required the newest version of a peer dependency, so the only solution is not to update.

The thing is that managing peer dependency versions is a complicated thing, and IMO the best way is to gradually drop supported versions from the end and release those in separate releases.

> if you are going to make sure every future version works with stylelint, why don't you just use `>=`?

 Because I don't know if the version that we have now works with a stylelint version that gets released two years from now. That's why the versions are whitelisted.

XhmikosR: Hence why I reverted that change and I just added v13.0.0 in peerDependencies.

I still believe this should be cleaned up later.

kristerkari: Yes, it definitely should be cleaned up, but I need to see if there is any data about downloads per stylelint version to get an idea if people are still using those old versions.

XhmikosR: Not sure where you can find such metrics TBH.

Anyway, I assume this PR should be good to go as is? I need to rebase a few of my other PRs after this gets merged.

package.json
libinvarghese: Can we use the [range](https://docs.npmjs.com/misc/semver#ranges) operator `>=`?

```
"stylelint": ">=8.0.0"
```

package.json
libinvarghese: I like the idea @XhmikosR suggested,
> if you are going to make sure every future version works with stylelint, why don't you just use >=

PR #434, opened 2020-01-14
Remove the unused github-contributors-list

XhmikosR:
@kristerkari if you want to add a script instead of removing the devDependency that's OK. But currently it's moot having it :) Also the current contributors are not up to date AFAICT.


kristerkari:
Thanks, I decided to add the script for now. It's still a half manual process, since I need to copy/paste the html that it generates.

PR #435, opened 2020-01-14
Stop using private Stylelint code

XhmikosR:
Split from #428 

**Should probably be merged after #428 and a rebase.**


XhmikosR:
Would be nice if we updated the other upstream code too and mark any custom code as such later.

XhmikosR:
@kristerkari this should have been merged after #428 and after close review. I mean, tests do pass, but you know

kristerkari:
@XhmikosR I thought that those Pull Requests don't depend on each other. Isn't this just a backwards compatible change anyway?

PR #442, opened 2020-01-16
Allow keyword value in any conditions

### LINKED ISSUE #438: Allow keywords in media-feature-value-dollar-variable
```scss
// valid
@media screen and (pointer: fine) and (min-width: $breakpoint) {}

// invalid
@media screen and (prefers-color-scheme: dark) and (min-width: 768px) {}
```
In my project, I want all media feature use variables for sizes. But I don't want to define variables for standard keywords like `fine`, `dark`.

guoyunhe:
Fix #438

CSS media features also have some keyword values, like `none`, `fine`, `srgb`. Usually you should not define a variable for them which is just overkill.

```scss
// valid
@media screen and (pointer: fine) and (min-width: $breakpoint) {}
```
In my project, I want all media feature use variables for sizes. But I don't want to define variables for standard keywords like `fine`, `dark`.

### DIFF
--- a/src/rules/media-feature-value-dollar-variable/index.js
+++ b/src/rules/media-feature-value-dollar-variable/index.js
@@ -1,5 +1,5 @@
 import { utils } from "stylelint";
-import { namespace } from "../../utils";
+import { namespace, optionsHaveIgnored } from "../../utils";
 
 export const ruleName = namespace("media-feature-value-dollar-variable");
 
@@ -9,12 +9,23 @@ export const messages = utils.ruleMessages(ruleName, {
     "Expected a dollar-variable (e.g. $var) to be used as a media feature value"
 });
 
-export default function(expectation) {
+export default function(expectation, options) {
   return (root, result) => {
-    const validOptions = utils.validateOptions(result, ruleName, {
-      actual: expectation,
-      possible: ["always", "never"]
-    });
+    const validOptions = utils.validateOptions(
+      result,
+      ruleName,
+      {
+        actual: expectation,
+        possible: ["always", "never"]
+      },
+      {
+        actual: options,
+        possible: {
+          ignore: ["keywords"]
+        },
+        optional: true
+      }
+    );
 
     if (!validOptions) {
       return;
@@ -29,6 +40,8 @@ export default function(expectation) {
     const variableRegex = /^\$[A-Za-z_0-9-]+$/;
     // `#{$var-name_sth}`
     const interpolationVarRegex = /^#\{\s*?\$[A-Za-z_0-9]+\s*?\}$/;
+    // `none`, `dark`
+    const keywordValueRegex = /^[a-z][a-z0-9-]*$/;
 
     root.walkAtRules("media", atRule => {
       const found = atRule.params.match(valueRegexGlobal);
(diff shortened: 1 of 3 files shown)


kristerkari:
Thanks @guoyunhe! 

Could you add tests for more cases than just the pointer? 

I'm also thinking that does this change break any existing cases that have maybe not been covered by the tests?

src/rules/media-feature-value-dollar-variable/index.js
```diff
@@ -53,6 +55,11 @@ export default function(expectation) {
           });
         }
 
+        // A keyword value, like `none`, should always be fine
+        if (valueParsed.search(keywordValueRegex) !== -1) {
```
XhmikosR: Maybe use `test` since you only check for a boolean here?

kristerkari:
Also, please add a test where you mix both cases, the one that needs a variable and the one that does not.

src/rules/media-feature-value-dollar-variable/index.js
guoyunhe: Thanks, update!

guoyunhe:
Added more tests with combined values, different features and keywords with dash and number.

This change makes the rule less strict so I think it will not break any existing configuration.

kristerkari:
Thanks for adding the tests @guoyunhe. I was thinking about this, and it's basically a breaking change to the rule for people who already have it in use. 

What if instead we would add a `ignore: ["keywords"]` option to the rule to enable this functionality?

guoyunhe:
For existing configuration people are using, they cannot use keyword value with this rule enabled. Otherwise stylelint will always throw errors whether they use "always" or "never". So it won't require users to do anything to migrate.

Since almost nobody would define keyword variables like: `$dark = dark;`, I would suggest to make it enabled by default and no need to opt-out/opt-in.

kristerkari:
> For existing configuration people are using, they cannot use keyword value with this rule enabled. Otherwise stylelint will always throw errors whether they use "always" or "never". So it won't require users to do anything to migrate.
> 
> Since almost nobody would define keyword variables like: `$dark = dark;`, I would suggest to make it enabled by default and no need to opt-out/opt-in.

I understand that what you described is something that the majority of the users will do, but since the rule did not consider media query keywords before, there are probably some people who have done exactly `$dark = dark;`, so the change needs to be considered a breaking change as we simply don't know how the users are using the rule.

guoyunhe:
If they use `$dark = dark;`, the code should be still valid. We don't introduce new errors.

But for safety, this can be marked as a breaking change or given opt-in/opt-out option. I will try to add opt-in option.

kristerkari:
> If they use `$dark = dark;`, the code should be still valid. We don't introduce new errors.

I know, but the rule will work differently, so a user might end up having some places in the code where a variable is used in place of a keyword, and some places where the keyword is used directly. 

That can happen without the developers realizing it if we don't have this change behind an option. That is the reason why making an existing rule less strict should be behind an option. Making the code more consistent is on of the reasons why people are using the linting rules.

guoyunhe:
Updated. By default, the rule behaves the same as before, until you specify `ignore: ["keywords"]`. Will update the README later.

kristerkari:
Great, thanks!

guoyunhe:
README updated.

PR #453, opened 2020-02-08
no-duplicate-dollar-variables: allow dup vars in unrelated scopes

### LINKED ISSUE #277: no-duplicate-dollar-variables doesn't respect scopes
`no-duplicate-dollar-variables` rule reports variables with the same name but declared in different scopes. The snippet below produces error `Unexpected duplicate dollar variable $padding-x   scss/no-duplicate-dollar-variables`.

```
.FormControl-label {
  $padding-x: $padding-small-x;

  @include input-overlay;

  position: relative;
  top: 0.40em;
  margin-left: $margin-medium-x;
  padding-left: $padding-x;
  padding-right: $padding-x;
  font-size: $font-size-small;
}


.FormControl-input {
  $padding-x: $padding-medium-x;

  width: 100%;
  border: 1px solid $color-grey-light;
  border-radius: 0;
  padding-left: $padding-x;
  padding-right: $padding-x;
  height: $row-height;
}
```

For me it's pretty obvious that this shouldn't be reported. Am I wrong?

srawlins:
Fixes #277

### DIFF
--- a/src/rules/no-duplicate-dollar-variables/index.js
+++ b/src/rules/no-duplicate-dollar-variables/index.js
@@ -32,6 +32,41 @@ export default function(value, secondaryOptions) {
 
     const vars = {};
 
+    /**
+     * Traverse the [vars] tree through the path defined by [ancestors], creating nodes as needed.
+     *
+     * Return the tree of the node defined by the last of [ancestors].
+     */
+    function getScope(ancestors) {
+      let scope = vars;
+
+      for (const node of ancestors) {
+        if (!(node in scope)) {
+          scope[node] = {};
+        }
+
+        scope = scope[node];
+      }
+
+      return scope;
+    }
+
+    /**
+     * Returns whether [variable] is declared anywhere in the scopes along the path defined by
+     * [ancestors].
+     */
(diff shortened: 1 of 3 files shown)


src/rules/no-duplicate-dollar-variables/index.js
```diff
@@ -60,7 +95,19 @@ export default function(value, secondaryOptions) {
         return;
       }
 
-      if (vars[decl.prop]) {
+      const ancestors = [];
+      let parent = decl.parent;
+
+      while (parent !== null && parent !== undefined) {
+        const parentKey = parent.toString();
```
srawlins: This seems to be inefficient - I think this serializes the entire subtree below [parent]. I'm not sure what to use as unique keys to traverse down the scopes though. the non-leaf nodes are all [Container](https://api.postcss.org/Container.html)s I think? I couldn't find a good common way to uniquify them...

kristerkari:
Thanks! This seems like a good fix to the rule.

srawlins:
Done. Additional test and README text.

kristerkari:
Great, thanks! Going to do a new release with the currently approved PRs.

PR #454, opened 2020-02-08
Do not find operators in valid unquoted URLs in url(); fixes #452

### LINKED ISSUE #452: scss/operator-no-unspaced examines unquoted urls
This test, added to `src/rules/operator-no-unspaced/__tests__/index.js` under `accept:`, at line 1263, fails:

```js
    {
      code: `
      @import url(//fonts.googleapis.com/css?family=Google+Material+Icons);
      `,
      description: "ignores unquoted urls"
    }
```

It complains about the `/` and `+` characters as operators.

srawlins:


kristerkari:
Thanks, and sorry for the delay. I have been on holiday.

src/rules/operator-no-unspaced/__tests__/index.js
```diff
@@ -2493,6 +2493,23 @@ testRule(rule, {
   syntax: "scss",
   skipBasicChecks: true,
 
+  accept: [
+    {
+      code: "@import url('10px +2');",
```
kristerkari: Could this be changed to be an actual valid url instead of just adding some math inside of a string?

srawlins: Done (note the existing reject cases include `url(10px +2)`).

PR #461, opened 2020-03-25
"scalar-color" should be "scale-color" in "function-color-relative" example

EvanHahn:
This appears to be a typo.


kristerkari:
Thanks!

Issue #361, opened 2019-08-23
comment-no-empty for scss


vankop:
Regarding to https://github.com/stylelint/stylelint/pull/4216 it will be great, if it will be rule `scss/comment-no-empty` for SCSS-like comments

kristerkari:
Implemented in #464

PR #464, opened 2020-04-07
New rule: comment-no-empty (#361)

airyboy:


src/rules/comment-no-empty/index.js
```diff
@@ -0,0 +1,37 @@
+import { utils } from "stylelint";
+import { namespace } from "../../utils";
+
+export const ruleName = namespace("comment-no-empty");
+
+export const messages = utils.ruleMessages(ruleName, {
+  expected: "Empty comments are forbidden"
```
kristerkari, line 7: Can you please make this match the rule from `stylelint`. So `rejected` instead of `expected` and change the error message to match.
https://github.com/stylelint/stylelint/blob/master/lib/rules/comment-no-empty/index.js#L9-L11

src/rules/comment-no-empty/__tests__/index.js
```diff
@@ -0,0 +1,75 @@
...
+    }    
+  ],
+  reject: [
+    {
+      code: `
+      /**/
+    `,
+      description: "Empty block comment",
+      message: messages.expected,
+      line: 2
```
kristerkari, line 29: Please also add column number to the tests.

kristerkari:
Looks good, but I had a couple of review comments that should be fixed before this can be merged. :)

airyboy:
Thank you, pushed a commit with fixes and added a couple tests as well

PR #469, opened 2020-05-06
Update devDependencies and drop Node.js < 10 support

XhmikosR:
@kristerkari feel free to push to this branch before merging to fix the remaining lint issues and ping me. Note that this PR targets the v4 branch.

Non-whitespace diff: <https://github.com/kristerkari/stylelint-scss/pull/469/files?w=1>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,8 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # switch to `8` when https://github.com/actions/setup-node/issues/27 is fixed
-        node: [8.16.2, 10, 12, 14]
+        node: [10, 12, 14]
         os: [ubuntu-latest, windows-latest]
 
     steps:
@@ -32,15 +31,15 @@ jobs:
 
       - name: Run jest tests
         run: npm run jest -- --runInBand
-        if: "!(startsWith(matrix.os, 'ubuntu') && matrix.node == 10)"
+        if: "!(startsWith(matrix.os, 'ubuntu') && matrix.node == 12)"
 
       - name: Run jest tests with coverage
         run: npm run jest -- --runInBand --coverage
-        if: startsWith(matrix.os, 'ubuntu') && matrix.node == 10
+        if: startsWith(matrix.os, 'ubuntu') && matrix.node == 12
 
       - name: Run Coveralls
         uses: coverallsapp/github-action@master
-        if: startsWith(matrix.os, 'ubuntu') && matrix.node == 10
+        if: startsWith(matrix.os, 'ubuntu') && matrix.node == 12
         with:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
           path-to-lcov: "./.coverage/lcov.info"
--- a/package.json
+++ b/package.json
@@ -15,12 +15,12 @@
   },
   "devDependencies": {
     "eslint": "^6.8.0",
-    "eslint-config-stylelint": "^11.1.0",
+    "eslint-config-stylelint": "^12.0.0",
     "eslint-plugin-lodash": "^7.1.0",
     "github-contributors-list": "^1.2.4",
     "husky": "^4.2.5",
-    "jest": "^25.5.4",
-    "jest-cli": "^25.5.4",
+    "jest": "^26.0.1",
+    "jest-cli": "^26.0.1",
     "lint-staged": "^10.2.2",
     "npmpub": "^5.0.0",
     "postcss": "^7.0.29",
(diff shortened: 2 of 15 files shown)


XhmikosR:
I disabled the jest-setup.js errors. There are only 2 left in `/src/utils/__tests__/parseFunctionArguments.js`

package.json
```diff
@@ -32,61 +32,31 @@
     "stylelint": "^8.0.0 || ^9.0.0 || ^10.0.0 || ^11.0.0 || ^12.0.0 || ^13.0.0"
   },
   "engines": {
-    "node": ">=8"
+    "node": ">=10"
```
XhmikosR, line 35: BTW if you want to make this stricter let me know. We should also change the version we test on CI to match the minimum one.

package.json
```diff
@@ -32,61 +32,31 @@
...
-      "template-curly-spacing": "error",
       "lodash/callback-binding": "error",
       "lodash/collection-method-value": "error",
       "lodash/collection-return": "error",
       "lodash/no-double-unwrap": "error",
       "lodash/no-extra-args": "error",
       "lodash/no-unbound-this": "error",
       "lodash/unwrap": "error",
-      "lodash/preferred-alias": "error",
-      "node/no-unsupported-features/es-syntax": [
```
XhmikosR, line 81: This is inherited by the stylelint eslint config. We could keep it and adapt it if you prefer that.

package.json
kristerkari: I think that this is good enough.

package.json
kristerkari: Sorry, what do you mean with "this"?

XhmikosR: This rule.

kristerkari:
> I disabled the jest-setup.js errors. There are only 2 left in `/src/utils/__tests__/parseFunctionArguments.js`

Yeah I can fix the current warnings/errors.

package.json
kristerkari: maybe we should keep it if we want to have a separate minimum node version from stylelint. It should not be a big problem if we inherit it, but we could run into some compatibility issues at some point.

XhmikosR:
@kristerkari maybe we should silence the jest/no-standalone-expect errors in jest-setup.js?

Feel free to fix those too. And also, if you want me to cherry pick any of these patches in master let me know. I can rebase this branch before the final merge.

package.json
XhmikosR: I'll have a look later an restore this.

XhmikosR:
Oh, and also, it would be nice if we could find a way to fix these warnings too :)

>You did not set any plugins, parser, or stringifier. Right now, PostCSS does nothing. Pick plugins for your case on https://www.postcss.parts/ and use them in postcss.config.js.

kristerkari:
> maybe we should **not** silence the jest/no-standalone-expect errors in jest-setup.js?

I originally took the code from stylelint's repo and did not want to change it, so that's why it's there. I haven't checked if their test runner code has changed since then.

kristerkari:
> Oh, and also, it would be nice if we could find a way to fix these warnings too

Those tests were missing the `parser` setting for PostCSS. Fixed now.

XhmikosR:
I think we could cherry pick a couple of patches like the postcss warnings fix and the missing it block patches. Do you want me to do sherry pick those to master? If so, I'll rebase this branch after that.

kristerkari:
> Do you want me to do sherry pick those to master? If so, I'll rebase this branch after that.

Please do :)

XhmikosR:
Done and branches rebased. Just make sure you don't override them.

If you want to release a new patch from master, now it's a good time. Just make sure you do `npm i` after you switch to master, and then your usual steps.

kristerkari:
Alright, I'll publish a patch with the changes in master.

XhmikosR:
Added back the `node/no-unsupported-features` rules with the version set to `10.0.0`.

I think we can merge this in the v4-dev branch if you agree.

.github/workflows/test.yml
```diff
@@ -11,8 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # switch to `8` when https://github.com/actions/setup-node/issues/27 is fixed
-        node: [8.16.2, 10, 12, 14]
+        node: [10, 12, 14]
```
XhmikosR, line 14: One last thing @kristerkari: this is the latest 10.x version. The same goes for the package.json engines version. I guess it's OK since we have the ESLint rules in place, but it's your call.

kristerkari: That is ok. We shouldn't need to care about minor versions.

PR #470, opened 2020-05-09
Use jest-preset-stylelint

hudochenkov:
We at stylelint released new version of [`jest-preset-stylelint`](https://github.com/stylelint/jest-preset-stylelint). This preset now comes with a `testRule` that we use in stylelint core to test all rules. You have old copy of that `testRule` in your [`jest-setup.js`](https://github.com/kristerkari/stylelint-scss/blob/27e644bdb1c1d7bf8ef64b79d2a07f2f81abe81a/jest-setup.js) (probably copied from `stylelint-order` :)).

New `testRule` has more checks, that the copy you have, therefore when use it right now 114 tests are failed.

There are three main reasons. All of which, when resolved in your tests, will give you more confidence.

1. All `reject` cases should have `message` specified.
2. If test case reports multiple violations all of them should be in `warnings` array.

	```diff
	{
		// ...
	-	message: '... 1',
	-	line: 1,
	-	column: 1,
	+	warnings: [
	+		{
	+				message: '... 1',
	+				line: 1,
	+				column: 1,
	+		},
	+		{
	+				message: '... 2',
	+				line: 2,
	+				column: 2,
	+		},
	+	]
	}
	```

3. Some test cases with `fix: true` doesn't apply fixes, because it's ignored, or `disableFix: true` is set for the rule. With old `testRule` workaround was to copy `code` to `fixed`. With new `testRule` `fixed` should be removed, and `unfixable: true` added:

	```diff
	{
	+	unfixable: true,
		code: `a {
		@if ($x == 1) {

		} @else {} @include x;
	}`,
	-	fixed: `a {
	-	@if ($x == 1) {
	-
	-	} @else {} @include x;
	- }`,
		description:
			"always-last-in-chain (followed by non-@else at-rule, no newline after).",
		message: messages.expected,
		line: 4
	}
	```

This PR is to show you setup with `jest-preset-stylelint`. After all tests are fixed `jest-setup.js` should be updated and all `testRule` should remove `rule` as the first argument:

```diff
- const testRule = getTestRule(stylelint, { plugins: ["./src"] });
- 
- global.testRule = (rule, schema) => {
-   testRule(schema);
- };
+ global.testRule = getTestRule(stylelint, { plugins: ["./src"] });
```


kristerkari:
Oh nice, I was wondering earlier if we could get rid of the code in `jest-setup.js`. The improvements to the tool seem really good.

I need to reserve some time to migrate all the tests that we have. Of course if someone wants to do it before that, help is always welcome :)

niksy:
I think this can be closed since commit has been cherry-picked to https://github.com/kristerkari/stylelint-scss/pull/543 ?

Issue #472, opened 2020-05-19
selector-no-redundant-nesting-selector rule bug


kaysonwu:
```less
.breadcrumb {
  // For less, the `&` here is not redundant
  & > span:last-child &-separator {
    display: none;
  }
}
```

Weird results will appear after removal:

```css
> span:last-child .breadcrumb-separator
```

Then it doesn't support `& when()` syntax:

```less
button {
  &:hover {
    // For less, the `&` here is not redundant
    & when (@theme = dark) {
     }

     // For less, the `&` here is not redundant
     & when not (@theme = dark) {
     }
  }
}
```

kaysonwu:
@kristerkari Hi, you need to enhance the selector-no-redundant-nesting-selector rule to better support `less`

kaysonwu:
@kristerkari I have prepared PR for this bug, please review it.

PR #473, opened 2020-05-19
fix less bug for #472

kaysonwu:
fix less bug for [#472](https://github.com/kristerkari/stylelint-scss/issues/472)


kaysonwu:
@kristerkari

src/rules/selector-no-redundant-nesting-selector/README.md
```diff
@@ -59,3 +59,51 @@ p {
...
 ```
+
+## Options
+
+Some false alarms can be blocked by assigning keywords:
+
+```js
+{
+  rules: {
+    'scss/selector-no-redundant-nesting-selector', [true, { keywords: ['when'] }],
```
kristerkari, line 70: The standard way to name the option would be to call it `ignoreKeywords`. But that makes me think that is `keyword` the correct term for this? Is that what it's called in Less?

src/rules/selector-no-redundant-nesting-selector/README.md
```diff
@@ -59,3 +59,51 @@ p {
   }
 }
 ```
+
+## Options
```
kristerkari, line 63: This should follow the formatting for options that is used in `stylelint` and other `stylelint-scss` rules, e.g. https://github.com/kristerkari/stylelint-scss/blob/master/src/rules/at-function-named-arguments/README.md#options

src/rules/selector-no-redundant-nesting-selector/README.md
kaysonwu: I took it to mean reserved. I can follow your principle if you think it necessary

src/rules/selector-no-redundant-nesting-selector/README.md
```diff
@@ -59,3 +59,51 @@ p {
...
+
+Some false alarms can be blocked by assigning keywords:
```
kristerkari, line 65: This is not a good way of describing the option. 

A better way would be to say something like: if you are using Less or some other non-SCSS syntax, the warnings can be disabled by using this option.

src/rules/selector-no-redundant-nesting-selector/README.md
```diff
@@ -59,3 +59,51 @@ p {
...
+
+```js
+{
+  rules: {
+    'scss/selector-no-redundant-nesting-selector', [true, { keywords: ['when'] }],
+  },
+}
+```
+
+The following patterns are *not* considered warnings:
```
kristerkari, line 75: There could be some kind of additional text here to say that this is Less syntax.

src/rules/selector-no-redundant-nesting-selector/__tests__/index.js
```diff
@@ -345,6 +345,50 @@ testRule(rule, {
...
+    },
+    {
+      code: `
+      @theme: ~'dark';
+      p {
+        & when (@theme = dark) {}
+        & when not (@theme = dark) {}
+      }
+    `,
+      line: 4,
```
kristerkari, line 357: Also add the `column` property here.

src/rules/selector-no-redundant-nesting-selector/README.md
kristerkari: Please have a look at this guide. It explains how to name options in stylelint rules:
https://stylelint.io/user-guide/rules/about#options

kaysonwu:
I've received your suggestion. I'm dealing with it

kaysonwu:
@kristerkari You need to review.

src/rules/selector-no-redundant-nesting-selector/__tests__/index.js
kristerkari: @kaysonwu can you still fix this one? The `column` property is missing from the test.

kaysonwu: I've fixed it, but it's not synced. I'll check it

kaysonwu: Now, it seems to be synchronized, I added the `color` property

src/rules/selector-no-redundant-nesting-selector/index.js
```diff
@@ -1,27 +1,34 @@
...
 export const ruleName = namespace("selector-no-redundant-nesting-selector");
 
 export const messages = utils.ruleMessages(ruleName, {
   rejected: "Unnecessary nesting selector (&)"
 });
 
-export default function(actual) {
+export default function(actual, options) {
   return (root, result) => {
     const validOptions = utils.validateOptions(result, ruleName, { actual });
```
kristerkari, line 17: Oh sorry I totally missed it on the first review round, but the `utils.validateOptions` call should include rule's secondary options too.

Have a look at how it's done here:
https://github.com/stylelint/stylelint/blob/master/lib/rules/font-family-no-duplicate-names/index.js#L25-L36

The option could also accept a regex as a value like the rule I just linked.

src/rules/selector-no-redundant-nesting-selector/README.md
```diff
@@ -59,3 +59,57 @@ p {
   }
 }
 ```
+
+## Options
+
+`string[]`: `[keywords]`
```
kristerkari, line 65: Sorry for nitpicking, but this should be `ignoreKeywords: ["/regex/", /regex/, "string"]`. The rule could be updated to also support regex values.

Here's an example:
https://github.com/stylelint/stylelint/tree/master/lib/rules/font-family-no-duplicate-names#ignorefontfamilynames-regex-regex-string

kaysonwu: This is the right pick，I will update immediately, remember to wait for me

kaysonwu:
@kristerkari   Please review it again, I am waiting for the new version

kaysonwu:
@kristerkari When can you review my PR?

kaysonwu:
Never mind, is there a plan to release a new version?

kristerkari:
> Never mind, is there a plan to release a new version?

Yes, I'm just going to quickly see if I can include one more bug fix for the release. I'll release the new version asap.

Issue #439, opened 2020-01-16
comment-no-loud


Jastos:
comment-no-loud is not disallowing /* comments

`/* MOBILE NAV STYLES */`

`"scss/comment-no-loud": true,`

kristerkari:
I just tested and it does give a warning for your example. Do you have something misconfigured?

mdimovska:
I have experienced the same problem. Apart from this, the behavior of this rule was a bit weird in  other cases. You can find some examples below.
 
* **Example 1:**
With the following scss content (example1.scss):
![image](https://user-images.githubusercontent.com/6908242/73934759-65bccb80-48df-11ea-8fd3-e10e8bcece4d.png)
stylelint does not complain:
![image](https://user-images.githubusercontent.com/6908242/73934732-56d61900-48df-11ea-84d3-a322e4251ef7.png)

* **Example 2:**
With the following scss content (example2.scss):
![image](https://user-images.githubusercontent.com/6908242/73934941-bfbd9100-48df-11ea-9f05-ad460e05287b.png)
I got completely different lint errors from what I expected (it complains that I use `//` instead of `/**/`):
![image](https://user-images.githubusercontent.com/6908242/73935003-df54b980-48df-11ea-8089-1aa0bce4e672.png)
After replacing `/**/` with `//` as shown below, there were no lint errors:
![image](https://user-images.githubusercontent.com/6908242/73935056-03b09600-48e0-11ea-88dc-95cba2de1c26.png)

* **Example 3:** 
What is more weird is that when I transferred the `/**/` comments from example2.scss to example1.scss, stylelint did not complain anymore:
![image](https://user-images.githubusercontent.com/6908242/73935149-3490cb00-48e0-11ea-8e2e-cacc595bc5ea.png)
![image](https://user-images.githubusercontent.com/6908242/73935156-378bbb80-48e0-11ea-9786-c12d2d35c116.png)


I am using stylelint-config-sass-guidelines@7.0.0 (with stylelint-scss@3.14.2).

mdimovska:
I have experienced the same problem. Apart from this, the behavior of this rule was a bit weird in  other cases. You can find some examples below.
 
* **Example 1:**
With the following scss content (example1.scss):
![image](https://user-images.githubusercontent.com/6908242/73934759-65bccb80-48df-11ea-8fd3-e10e8bcece4d.png)
stylelint does not complain:
[...]

PR #475, opened 2020-05-25
Fix comment-no-load only covering first line

kevindew:
Addresses https://github.com/kristerkari/stylelint-scss/issues/439

This changes the approach to looking up comments for the comment-no-loud
rule. This rule had an issue where it would only identify comments if
they were the first node within an SCSS file.

Previously this check matched a regex against `source.input.css` of the
comment node. `source.input.css` is actually the full input of the
source file and not the input of the node (as I imagine was expected
when this was initially authored). This meant that only comments that
were not preceded by any other nodes matched the rule. The tests all
succeeded for this as they all checked against the first node.

This new approach instead takes the source input and splits into an array
based on a `\n` separator. It then looks up the first line of the
comment with `source.start.line` variable. This provides the first
line of the comment input as was given in the source file.

I did also try using `comment.toString()` as a simpler means to get the
original source value, however this approach didn't provide an exact
copy of the original input and instead converted SCSS comments (`//`)
into CSS ones (`/*`) meaning that all SCSS comments were identified as
loud.

### DIFF
--- a/src/rules/comment-no-loud/__tests__/index.js
+++ b/src/rules/comment-no-loud/__tests__/index.js
@@ -35,6 +35,23 @@ testRule(rule, {
       line: 2,
       column: 9,
       message: messages.expected
+    },
+    {
+      code: `
+      i {
+        font-style: italic;
+      }
+
+      /* comment line */
+
+      b {
+        font-weight: bold;
+      }
+    `,
+      description: "Comment sandwiched between rules",
+      line: 6,
+      column: 7,
+      message: messages.expected
     }
   ]
 });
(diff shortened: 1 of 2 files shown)


src/rules/comment-no-loud/__tests__/index.js
```diff
@@ -31,6 +31,21 @@ testRule(rule, {
...
+      }
+
+      /* comment line */
+
+      b {
+        font-weight: bold;
+      }
+    `,
+      description: "Comment sandwiched between rules",
+      message: messages.expected
```
kristerkari, line 48: I noticed that the existing tests are missing the `line` and `column` numbers from the rejected test cases. Could you please add those to your test?

kevindew: @kristerkari Sure, I've added those in now.

PR #481, opened 2020-07-06
Fix duplicate rule names in `at-rule-no-unknown` message

ybiquitous:
Fix #480

PR #484, opened 2020-07-08
Unify messages for `comment-no-empty` and `scss/comment-no-empty`

ybiquitous:
This change aims to respect the message of the stylelint's core rule [`comment-no-empty`](https://github.com/stylelint/stylelint/blob/a8042e7192cad993081cf982901948dbd5004048/lib/rules/comment-no-empty/index.js#L12).
Because the `scss/comment-no-empty` is essentially an extension of the core rule.

Related to #481 and #483

### DIFF
--- a/src/rules/comment-no-empty/index.js
+++ b/src/rules/comment-no-empty/index.js
@@ -6,11 +6,10 @@ const coreRuleName = "comment-no-empty";
 export const ruleName = namespace(coreRuleName);
 
 export const messages = utils.ruleMessages(ruleName, {
-  rejected: (...args) => {
-    return rules[coreRuleName].messages
-      .rejected(...args)
-      .replace(` (${coreRuleName})`, "");
-  }
+  rejected: rules[coreRuleName].messages.rejected.replace(
+    ` (${coreRuleName})`,
+    ""
+  )
 });
 
 function rule(primary) {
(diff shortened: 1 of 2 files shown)


src/rules/comment-no-empty/index.js
```diff
@@ -1,10 +1,16 @@
...
+import { rules, utils } from "stylelint";
 import { namespace } from "../../utils";
 
-export const ruleName = namespace("comment-no-empty");
+const coreRuleName = "comment-no-empty";
+
+export const ruleName = namespace(coreRuleName);
 
 export const messages = utils.ruleMessages(ruleName, {
-  rejected: "Unexpected empty comment"
```
ybiquitous, line 7: This message is the same as the core rule's one, so this PR is just a refactoring.

kristerkari:
@ybiquitous would it be possible to get a unit test for this one just like in #481?

ybiquitous:
@kristerkari I've pushed the change (8f66725) to fix the implementation and add a test case!

kristerkari (approved):
Looks good 👍

ybiquitous:
@kristerkari Thank you! 😄

kristerkari:
@ybiquitous Thank you for the contribution, and sorry about the long delay :)

ybiquitous:
No worries! 🙆🏼‍♂️

PR #491, opened 2020-09-20
Fix lack of note about autofix for at-rule-conditional-no-parentheses rule

chalkygames123:
It's not documented, but the rule seems to have the ability of autofix.


kristerkari:
Thanks!

Issue #500, opened 2020-11-23
[Request] `no-duplicate-dollar-variables` to take !default into consideration


joseph118:
Currently, the rule doesn't consider the default keyword and throws a lint error when the following scenario is present;
```
$a: () !default;
$a: map-merge($defaults, $a);
```

I believe that the rule should at least be configurable and considers duplicate with default keyword to an extent

kristerkari:
Closing as this was fixed in #502 👍

PR #502, opened 2020-11-23
add `ignoreDefaults` options on no duplicate dollar var rule

joseph118:
https://github.com/kristerkari/stylelint-scss/issues/500

something to start off with regards to the raised request (still needs to add documentation in this case and add more tests)

will continue once feedback is provided :)

### DIFF
--- a/src/rules/no-duplicate-dollar-variables/README.md
+++ b/src/rules/no-duplicate-dollar-variables/README.md
@@ -77,7 +77,9 @@ $a: 1;
   $b: 2;
 }
 ```
+
 ___
+
 ### `ignoreInside: ["at-rule", "nested-at-rule"]`
 
 #### `"at-rule"`
@@ -127,7 +129,9 @@ $a: 1;
   }
 }
 ```
+
 ___
+
 ### `ignoreInsideAtRules: ["array", "of", "at-rules"]`
 
 Ignores all variables that are inside specified at-rules.
@@ -155,7 +159,9 @@ $a: 1;
   $a: 2;
 }
 ```
+
 ___
+
 ### `ignoreDefaults: [boolean, number]`
 
 Ignores all or an amount of variables containing the `!default` keyword.
--- a/src/rules/no-duplicate-dollar-variables/README.md
+++ b/src/rules/no-duplicate-dollar-variables/README.md
@@ -24,6 +24,22 @@ A dollar variable is considered a duplicate if it shadows a variable of the same
 }
 ```
 
+A dollar variable is **not** considered a duplicate if it contains the `!default` keyword (see the [Sass documentation](https://sass-lang.com/documentation/variables#default-values)). Two dollar variables are duplicates if they both contain `!default` keyword.
+
+```scss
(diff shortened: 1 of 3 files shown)


kristerkari:
Thanks, and sorry for the delay. I have been very busy lately. I'll try to get this reviewed asap.

kristerkari:
I think that this looks like a good option to add, but I'm unsure about the name. Somehow `markDefaults` seems like a bit confusing name to me. @joseph118 would you have any alternative names in mind?

joseph118:
I wasn't sure what to name it @kristerkari, but it does sound a bit confusing. Perhaps `ignoreDefaults`?

____

As of now, it isn't ignoring all defaults but just the first one, and I think that we should adjust it to be able to ignore all default vars or allow this possibility.
What if `ignoreDefaults` can accept either a boolean or an integer value, which will be used to either ignore the defaults 'n' times or all the time.

'n' would be the number given where if `1` is given, it will ignore it once but throws an error if a 2nd default with the same name is found

joseph118:
@kristerkari improved existing logic, added documentation, and tests.

Should be all good now 👍

kristerkari:
> What if `ignoreDefaults` can accept either a boolean or an integer value, which will be used to either ignore the defaults 'n' times or all the time.
> 
> 'n' would be the number given where if `1` is given, it will ignore it once but throws an error if a 2nd default with the same name is found

I haven't really been using the `!default` flag, so in what situation would you have multiple variables of the same name using `!default`. Can you give some example? I'm asking because I want to understand how useful the number flag is.

joseph118:
From my perspective, there are 2 common situations that are definitely going to be used;

- ignore all `!default` keyword (handled via boolean - `true`)
- and warning for duplicate declarations (handled via number - `1`)

____

Now for the other inputs (`1 < number`), this was only added to allow flexibility for the user but can be discarded just by allowing possibility of - boolean or `1`

kristerkari:
@joseph118 What I was after was that do you ever need to use multiple `!default` keywords for a variable? If not, then could the behavior of numer `1` be used by default and not giving any number option? As I said before, I haven't been using `!default` before, so I'm not sure how it's used.

joseph118:
@kristerkari In theory no, it doesn't make sense to have multiple `default` keywords for the same variable even though they are ignored if the variable is already set.

In a nutshell, when the `default` keyword is defined next to a variable, sass would check if this variable is declared before setting it and if it is declared, it will just ignore it without doing anything. (Basically don't override if it is declared)
___
That does make sense but I did it this way to avoid breaking changes since the behavior is going to change as `default` will be a var on its own internally now.

joseph118:
@kristerkari I've gone through it again and removed the possibility of providing a number but can still provide a boolean turn off/on the option. 
Let me know if you would like any more adjustments 👍

joseph118:
🏓 pinging @kristerkari

kristerkari:
Sorry, been busy lately, but this is good to be merged and released in a new version. Thanks!

PR #512, opened 2021-02-18
Add at-import-no-import rule

### LINKED ISSUE #510: Rule idea: Disallow @import
According to the [Sass documentation](https://sass-lang.com/documentation/at-rules/import), you should no longer use the `@import` rule:

> The Sass team discourages the continued use of the `@import` rule. Sass will gradually phase it out over the next few years, and eventually remove it from the language entirely. Prefer the `@use` rule instead.

It would be very useful to have a rule, which disallows the usage of the `@import` rule.

json-derulo:
Closes #510

### DIFF
--- a/src/rules/at-import-no-import/index.js
+++ b/src/rules/at-import-no-import/index.js
@@ -1,0 +1,29 @@
+import { utils } from "stylelint";
+import { namespace } from "../../utils";
+
+export const ruleName = namespace("at-import-no-import");
+
+export const messages = utils.ruleMessages(ruleName, {
+  expected: "@import is disallowed, replace it with @use"
+});
+
+export default function(on) {
+  return (root, result) => {
+    const validOptions = utils.validateOptions(result, ruleName, {
+      actual: on
+    });
+
+    if (!validOptions) {
+      return;
+    }
+
+    root.walkAtRules("import", node =>
+      utils.report({
+        message: messages.rejected,
+        node,
+        result,
+        ruleName
+      })
+    );
+  };
+}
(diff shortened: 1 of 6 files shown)


kristerkari:
Thanks! I'll try to review asap.

src/rules/at-import-no-import/index.js
```diff
@@ -0,0 +1,29 @@
+import { utils } from "stylelint";
+import { namespace } from "../../utils";
+
+export const ruleName = namespace("at-import-no-import");
+
+export const messages = utils.ruleMessages(ruleName, {
+  expected: "@import is disallowed, replace it with @use"
```
kristerkari, line 7: Could you please change the validation message to match the stylelint format:
https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules/about.md#violation-messages

json-derulo:
@kristerkari fixed the rule name, anything else?

kristerkari:
Thanks for the fix and sorry for the delay @json-derulo

kristerkari:
Hmm.. I'm thinking about the name, would it be clearer if the rule was just called `no-at-import`? Just like there is already a rule called `no-dollar-variables`.

json-derulo:
@kristerkari agree to that, changed the rule name.

kristerkari:
@json-derulo nice! I Think that this is ready then!

kristerkari:
@json-derulo Thanks for all the work you put into this rule! I should have checked earlier if stylelint already has something similar.

PR #519, opened 2021-04-22
Update README to move Rules to independent doc

blackfalcon:
This pull request aims to improve documentation. By moving the rules to an independent file, it's easier to draw links to this document and reference content.


kristerkari:
I kind of like that the npm page show the readme with the rules, so I'm not sure if I agree about the improvement, but I'm open to hear if there is something else that makes this a better way.

blackfalcon:
> I kind of like that the npm page show the readme with the rules, so I'm not sure if I agree about the improvement, but I'm open to hear if there is something else that makes this a better way.

I appreciate your opinion on this. My point is that it's not customary to place all documentation in the README with all other repo information. Placing it into another document resource allows for updates to documentation per the tool not to be muddied with updates to documentation about the repo. This also makes linking to the tool's docs more direct versus linking to the middle of a README. 

The most relative example would be [stylelint](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configure.md)

kristerkari:
> I appreciate your opinion on this. My point is that it's not customary to place all documentation in the README with all other repo information. Placing it into another document resource allows for updates to documentation per the tool not to be muddied with updates to documentation about the repo. This also makes linking to the tool's docs more direct versus linking to the middle of a README.

Thanks for your input!

I know that it's quite common to separate the README to smaller parts. However, I don't think that the `stylelint-scss` README is yet so big that you could not find the information that you need easily. What comes to the stylelint README: I find that it's usually difficult for me to navigate from the front page to the list of rules or the developer docs.

I understand that linking to the rules is  better when they are on a separate page, so I need to think about this a bit more.

blackfalcon:
I will throw in a personal agenda as well. I am building a design system and part of that is using tools like yours to maintain quality. I'd like to showcase the rules in my doc site much as I do with [stylelint](http://auro.alaskaair.com/css/linter), but I'd prefer not to have to maintain this myself or have to show the whole README with the rules.

stof:
@blackfalcon the list of _available_ rules in the plugin does not mean that this is the list of rules applied in your design system. And each rule is already linkable by being documented in its own readme, so you could link to the rules you enable.

blackfalcon:
> @blackfalcon the list of _available_ rules in the plugin does not mean that this is the list of rules applied in your design system. And each rule is already linkable by being documented in its own readme, so you could link to the rules you enable.

I'd argue that more often than not that all the rules are enabled at first and then paired down individually if needed. The goal I am striving for is to have a centralized list of all the features so that your developers can evaluate as needed. 

Having this all in the README is just making them parse through additional content to get what they are looking for. 

But this is your repo and if you don't agree with that, please feel free to close my PR.

XhmikosR:
Closing this per @kristerkari's comments above.

PR #532, opened 2021-07-14
Eslint-like examples for dollar-variable-pattern

SterlingVix:
Provide better examples and formatting / usage instructions in the form of comment blocks like ESLint. See: https://eslint.org/docs/rules/lines-between-class-members#options

Ensure example formatting is consistent throughout README doc.

### DIFF
--- a/src/rules/dollar-variable-pattern/README.md
+++ b/src/rules/dollar-variable-pattern/README.md
@@ -14,18 +14,34 @@ a { $foo: 1px; }
 
 A string will be translated into a RegExp like so `new RegExp(yourString)` — so be sure to escape properly.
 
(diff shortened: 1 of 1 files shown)


SterlingVix:
Hey all -- I was confused at the instructions for `dollar-variable-pattern` and saw some comments echoing my confusion, so I updated the docs.

I tried to preserve the standard stylelint docs formatting, but included examples formatted like ESLint rules examples. They contain both examples of working/breaking code *and* inline usage, which, I hope, will reduce confusion for other people trying to use this rule.

kristerkari:
@SterlingVix This is great! Thanks a lot! 

The updates to improve the READMEs of the rules are really important. 👍

PR #533, opened 2021-07-17
dollar-variable-first-in-block: support ignoring @use and @forward

### LINKED ISSUE #517: dollar-variable-first-in-block - ignore @use
I think `"ignore": ["imports"]` option should also ignore `@use`.

Real-world example:

```scss
@use "sass:color";

$primary-color: #f26e21 !default;
$secondary-color: color.change($primary-color, $alpha: 0.08) !default;
```

Actual: `Expected $-variable to be first in block (scss/dollar-variable-first-in-block)`

Expected: no warnings

kristerkari:
Improve the `dollar-variable-first-in-block` rule's `ignore: ["imports"]` option so that it also ignores `@use` and `@forward`.

fixes #517

ping @EvgenyOrekhov

### DIFF
--- a/src/rules/dollar-variable-first-in-block/README.md
+++ b/src/rules/dollar-variable-first-in-block/README.md
@@ -9,7 +9,7 @@ Require `$`-variable declarations to be placed first in a block (root or a rule)
 The following patterns are considered violations:
 
 ```scss
-@import '1.css';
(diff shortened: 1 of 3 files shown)


EvgenyOrekhov (approved):
:+1:

Issue #541, opened 2021-09-17
Help us prepare for the next major release of stylelint


jeddy3:
We've [been working on](https://github.com/stylelint/stylelint/issues/5205) a new major release of stylelint. There are [quite a few changes](https://github.com/stylelint/stylelint/blob/v14/CHANGELOG.md#head) (and a [lot of refactoring](https://github.com/stylelint/stylelint/pulls?q=is%3Apr+refactor+)). One of the biggest changes we've made is [removing the `syntax` option](https://github.com/stylelint/stylelint/issues/5289#issuecomment-840772900) and no longer bundling syntaxes in stylelint itself. Users will need to extend stylelint (via shared-configs, plugins and custom syntaxes) to lint something other than vanilla CSS.

To make this easier for users to do, we've added two properties to the configuration object:
- [`customSyntax`](https://github.com/stylelint/stylelint/pull/5538) - which accepts an PostCSS syntax like [`postcss-scss`](https://github.com/postcss/postcss-scss)
- [`overrides`](https://github.com/stylelint/stylelint/pull/5521) - which allows users to configure stylelint to lint two or more languages, e.g. CSS and SCSS

We [intend to improve the migration experience](https://github.com/stylelint/stylelint/issues/5301) by updating our docs and by adding helpful error messages to stylelint when a user, for example, tries to lint an `*.scss` file but hasn't specified a custom syntax in their configuration object.

This last part is [why we're reaching out to you](https://github.com/stylelint/stylelint/issues/5205#issuecomment-856669471). We want to encourage users to extend a shared-config for the language they are linting, rather than add custom syntaxes and plugins themselves.

We also think emphasising the [`standard`](https://github.com/stylelint/stylelint-config-standard) configs over the [`recommended`](https://github.com/stylelint/stylelint-config-recommended) ones will help users, especially new ones. Including those who use stylelint and a pretty-printer like Prettier together, as there are a number of stylistic rules in stylelint and this plugin that are complementary to Prettier, e.g. the `*-empty-line-before` ones.

To help us do this, will you (or any of the other contributors to this package) be willing to?:

1. Update [`stylelint-config-recommended-scss`](https://github.com/kristerkari/stylelint-config-recommended-scss) to bundle [`postcss-scss`](https://github.com/postcss/postcss-scss)
2. Create and publish `stylelint-config-standard-scss` that would:
[...]

PR #543, opened 2021-09-27
Use Stylelint v14

niksy:
As per https://github.com/kristerkari/stylelint-scss/issues/541#issuecomment-927022891, I’ve made changes for Stylelint 14 support.

First, this PR contains everything from https://github.com/kristerkari/stylelint-scss/pull/470 with fixed tests and this works with version 13.

I’ve added Stylelint 14 support from v14 branch but I’m getting errors. @jeddy3, could you please look if I’m doing everything right?


src/utils/isCustomPropertySet.js
```diff
@@ -8,12 +7,13 @@ import hasBlock from "./hasBlock";
  * @returns {boolean}
  */
 export default function(node) {
-  const selector = get(node, "raws.selector.raw", node.selector);
```
niksy, line 11: @kristerkari this had to be modified since there’s a difference between how custom properties are parsed in PostCSS 7 and 8. They’re declarations instead of rules. Can’t see anything in changelog though :/

kristerkari:
Thanks @niksy for taking care of this! I will try to start reviewing this as soon as I have time for it

jeddy3:
@niksy Amazing work! Especially migrating to the new jest preset as it offers more robust testing of rules.

The node 8 and node 10 tests are failing because stylelint requires a minimum of node 12.20. You'll need to update your workflow to remove node 8 and 10 from the matrix. I also suggest using node 14 for the coveralls run.

Node 12 and 14 have 199 failed and 1911 passed tests. 

Can you try:
- replacing all instances of `config: [undefined]` with `config: [true]` as that may be throwing the preset off
- skipping the HTML and Markdown tests (use `skip: true` in the `testRule`) as the HTML and Markdown syntaxes aren't yet compatible with PostCSS 8
- convert these tests to use `testRule` as 1. multiple warnings can now be checked by the preset 2. Stylelint rules aren't standalone PostCSS 8 compatible plugins and so can't be used in that way
  - https://github.com/kristerkari/stylelint-scss/blob/master/src/rules/at-rule-no-unknown/__tests__/index.js#L205
  - https://github.com/kristerkari/stylelint-scss/blob/master/src/rules/operator-no-unspaced/__tests__/index.js#L1843
  - any other similar ones

Let's see what we're left with after that.

niksy:
@jeddy3 all done! It seems like tests are now passing.

src/utils/isCustomPropertySet.js
jeddy3: We ran into the same change in behaviour with migrating to PostCSS 8.

Your changes to the util LGTM.

.github/workflows/test.yml
```diff
@@ -11,8 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # switch to `8` when https://github.com/actions/setup-node/issues/27 is fixed
-        node: [8.16.2, 10, 12, 14]
+        node: [12.20, 14]
```
jeddy3, line 14: We can use `12` here as it'll use the latest version, which is greater than 12.20.

.github/workflows/test.yml
```diff
@@ -33,17 +32,12 @@ jobs:
       - name: Build
         run: npm run build
 
-      - name: Run jest tests
```
jeddy3, line 36: Coverage is slow. We may want to revert these changes, and simply `matrix.node == 10` to `matrix.node == 14` in the conditions.

.github/workflows/test.yml
```diff
@@ -33,17 +32,12 @@ jobs:
...
-        if: "!(startsWith(matrix.os, 'ubuntu') && matrix.node == 10)"
-
       - name: Run jest tests with coverage
         run: npm run jest -- --runInBand --coverage
-        if: startsWith(matrix.os, 'ubuntu') && matrix.node == 10
 
       - name: Run Coveralls
         uses: coverallsapp/github-action@master
-        if: startsWith(matrix.os, 'ubuntu') && matrix.node == 10
+        if: startsWith(matrix.os, 'ubuntu') && matrix.node == 16
```
jeddy3, line 40: We can use `14`, not `16` here.

jeddy3 (changes requested):
Fantastic stuff! You've done a tonne of great work here, especially migrating all those custom tests for multiple warnings to `testRule`.

I'm glad my suggestions solved all the failing tests. I was anticipating a few would be left because of breaking behavioural changes in the parser, but it looks like this plugin is unaffected.

> Should we maybe add 16 to matrix as well?

I think so. 

I've made a couple of suggestions for the workflow. Other than that, this pull request LGTM!

niksy:
@jeddy3 

> I'm glad my suggestions solved all the failing tests. I was anticipating a few would be left because of breaking behavioural changes in the parser, but it looks like this plugin is unaffected.

Actually there are 2 tests that are suspicious. In previous version they were producing one warning, but in new version there are no warnings. I think it’s related to nested property parsing. @kristerkari do you know more about this? I’ve [reverted](https://github.com/kristerkari/stylelint-scss/pull/543/commits/2e3b972c5654e0cb43e3b4e121ce9a8e44cf9413) those 2 tests with their respective warnings.

I’ve made all other changes and taken the time to [upgrade](https://github.com/kristerkari/stylelint-scss/pull/543/commits/b6609cf8a3aee63a8a75f49e0b776a07dd3c3ef1) Jest related packages to newer version.

niksy:
OK, I’ve managed to fix one test. It seems like PostCSS doesn’t parse nested declarations as rules so we need to check for [`NestedDeclaration` from `postcss-scss`](https://github.com/postcss/postcss-scss/blob/main/lib/nested-declaration.js).

This fixes one test, but one is left, and it doesn’t pass because source line and column of that declaration is something completely different from original source code. I think parsing in `postcss-scss` does some [weird thing with tokens](https://github.com/postcss/postcss-scss/blob/main/lib/scss-parser.js#L57) resulting in different line and column values.

niksy:
Fix was published upstream so all tests are now passing! 🥳

niksy:
As for Node 16 testing, one test is failing and it’s related to Less syntax parsing: https://github.com/niksy/stylelint-scss/blob/stylelint-v14/src/rules/selector-no-union-class-name/__tests__/index.js#L150-L178:

```
 FAIL  src/rules/selector-no-union-class-name/__tests__/index.js
  ● Test suite failed to run

    Jest worker encountered 4 child process exceptions, exceeding retry limit

      at ChildProcessWorker.initialize (node_modules/jest-worker/build/workers/ChildProcessWorker.js:193:21)
```

I think it’s something related to [error execution count in Jest](https://github.com/facebook/jest/issues/8769).

kristerkari:
> Sidenote, why do we handle Less stuff in plugin for Sass files?

It's nothing that is very mandatory, but there are some rules which are so simple that the same rule can be applied to Less. I'm totally find with dropping support for such things.

kristerkari:
Great work with the changes @niksy !

I think that after stylelint v14 gets released, we can create a 3.x branch from the current master and release these changes as stylelint-scss 4.0.0

niksy:
> I think that after stylelint v14 gets released, we can create a 3.x branch from the current master and release these changes as stylelint-scss 4.0.0

Great! I see there is already [v4 branch](https://github.com/kristerkari/stylelint-scss/tree/v4) which was created a year ago. Are there any changes that can be reused? Is that branch still relevant?

> It's nothing that is very mandatory, but there are some rules which are so simple that the same rule can be applied to Less. I'm totally find with dropping support for such things.

This one test rejects Promise but I don’t see how that can be handled with `testRule`. Other Less tests are passing.

kristerkari:
> Great! I see there is already [v4 branch](https://github.com/kristerkari/stylelint-scss/tree/v4) which was created a year ago. Are there any changes that can be reused? Is that branch still relevant?

No idea to be honest. There hasn't been any plans for v4, so that might be just created to introduce some breaking change. It's fine to just ignore that branch as irrelevant.

jeddy3 (approved):
Fantastic work! LGTM.

To give users the smoothest possible migration experience, it'd be great if `4.0.0` of this package was released at the same time as `14.0.0` of stylelint. I'm happy to do the coordinating (as such a large number of our users lint SCSS and we'll be pointing to the upcoming standard-scss config in our migration guide).

@kristerkari Would you be willing to add me as a contributor to the repo and the npm packages of:
- this plugin
- the recommended-scss config

@niksy And to the upcoming standard-scss config

It'll just be for the 14.0.0 release.

kristerkari:
> @kristerkari Shall we create a `3.x` branch and merge this into `master`? I can then create a follow-up pull request.

Yes, we should do that, create 3.x from the current master and then get this in.

jeddy3:
> Yes, we should do that, create 3.x from the current master and then get this in.

Excellent. I'll get on it.

PR #566, opened 2021-11-24
Allow using the CSS min and max calculations and the CSS filter functions

### LINKED ISSUE #486: no-global-function-names does not allow using CSS math functions
This code causes the error `Expected math.min instead of min   scss/no-global-function-names`

```scss
.foo {
  width: min(300px, 30%);
}
```

Same with the _max()_ function `Expected math.max instead of max   scss/no-global-function-names` 
```scss
.foo {
  width: max(300px, 30%);
}
```

.stylelintrc.js
```js
module.exports = {
  'plugins': [
    'stylelint-scss',
  ],
  'rules': {
    'scss/no-global-function-names': true,
  },
};
```

```json
"stylelint": "13.7.0",
"stylelint-scss": "3.18.0"
```

https://caniuse.com/#feat=css-math-functions

### LINKED ISSUE #560: False positive max and min in scss/no-global-function-names
Input:

```
div {
  min-width: max(30vw, 300px);
}
```

Require:

```
div {
  min-width: math.max(30vw, 300px);
}
```
Expected (no errors):

```
div {
  min-width: max(30vw, 300px);
}
```

### LINKED ISSUE #562: no-global-function-names doesn't allow css filter declaration
`no-global-function-names` shows error for `backdrop-filter: saturate(140%) blur(10px);` cause of the `saturate`.
I think there should be exceptions for filter and backdrop-filter.

stof:
The Sass functions should indeed be migrated to the sass:math module. But min() and max() can also be the CSS functions (and with first-class calc in Sass, it will parse it as a calculation if possible).

Closes #560
Closes #562
Closes #486

### DIFF
--- a/src/rules/no-global-function-names/__tests__/index.js
+++ b/src/rules/no-global-function-names/__tests__/index.js
@@ -40,6 +40,22 @@ testRule({
     },
     {
       code: `
+      a {
+       min-width: min(30vw, 300px);
+      }
+    `,
+      description: "An allowed global function"
+    },
+    {
+      code: `
+      a {
+       min-width: max(30vw, 300px);
+      }
+    `,
+      description: "An allowed global function"
+    },
+    {
+      code: `
       @use "sass:color";
       a {
        background: color.red(#6b717f);
--- a/src/rules/no-global-function-names/__tests__/index.js
+++ b/src/rules/no-global-function-names/__tests__/index.js
@@ -72,6 +72,14 @@ testRule({
     },
     {
       code: `
+      a {
+       filter: alpha(opacity=50);
+      }
+    `,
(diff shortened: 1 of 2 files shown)


kristerkari (approved):
Makes sense, we could just not lint those for now. 👍

kristerkari:
@stof could you please also add unit tests for this change.

stof:
@kristerkari done. and I also solved the case of `filter` that was added in a comment in the issue.

kristerkari:
@stof looks like some unit tests are failing

stof:
@kristerkari fixed

kristerkari:
Thanks for the contribution @stof !

PR #567, opened 2021-11-25
fix: use processSync

xboy2012:
in the Stylelint source code, it use processSync.

process() returns a Promise but it's not handled properly.

### DIFF
--- a/src/utils/parseSelector.js
+++ b/src/utils/parseSelector.js
@@ -2,7 +2,7 @@ import postCssSelectorParser from "postcss-selector-parser";
 
 export default function(selector, result, node, cb) {
   try {
-    postCssSelectorParser(cb).process(selector);
+    postCssSelectorParser(cb).processSync(selector);
   } catch (e) {
     result.warn("Cannot parse selector", { node });
   }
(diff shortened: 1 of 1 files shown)


xboy2012:
anyone come to review this?

src/utils/parseSelector.js
```diff
@@ -2,7 +2,7 @@ import postCssSelectorParser from "postcss-selector-parser";
 
 export default function(selector, result, node, cb) {
   try {
-    postCssSelectorParser(cb).process(selector);
+    postCssSelectorParser(cb).processSync(selector);
```
kristerkari, line 5: We could make this fully match the code from stylelint and make add a return: `return postCssSelectorParser(cb).processSync(selector)`

kristerkari: ping @xboy2012

xboy2012: Sorry for late response.

I was working on an old version of stylelint, in which I see no `return` in the source code.

I read the latest code of stylelint again. I see the `return` there. My PR is Updated.

xboy2012: For convenience, I post the source code url here

https://github.com/stylelint/stylelint/blob/060310cfd6383175ee06fe4d15aee27a4e202671/lib/utils/parseSelector.js

PR #569, opened 2021-12-01
Implement the at-use-no-unnamespaced rule

stof:
Closes https://github.com/stylelint-scss/stylelint-scss/issues/565

### DIFF
--- a/src/rules/at-use-no-unnamespaced/index.js
+++ b/src/rules/at-use-no-unnamespaced/index.js
@@ -1,0 +1,29 @@
+import { utils } from "stylelint";
+import { namespace } from "../../utils";
+
+export const ruleName = namespace("at-use-no-unnamespaced");
+
+export const messages = utils.ruleMessages(ruleName, {
+  rejected: "Unexpected @use without namespace"
+});
+
+export default function(actual) {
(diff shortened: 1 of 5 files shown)


stof:
My implementation is incomplete. It won't catch configured modules, as the configuration comes after the namespace in the rule.

stof:
And here is the complete implementation

src/rules/at-use-no-unnamespaced/index.js
```diff
@@ -0,0 +1,29 @@
+import { utils } from "stylelint";
+import { namespace } from "../../utils";
+
+export const ruleName = namespace("at-use-no-unnamespaced");
+
+export const messages = utils.ruleMessages(ruleName, {
+  expected: "Expected @use to use a namespace"
```
kristerkari, line 7: I went and checked from the stylelint repo, and most of the `-no-` rules seem to use the `rejected: "Unexpected ..."` message format, should this rule also use it?

stof: I updated it

kristerkari:
This was released in 4.1.0:

Thank you @stof !

PR #570, opened 2021-12-01
Implement the dollar-variable-no-namespaced-assignment rule

### LINKED ISSUE #556: Rule idea: forbid assignment to namespaced variables
When using a Sass module with `@use`, exported variables are writable currently (except for built-in modules, so you cannot reassign `math.$pi` to a different value).
I started a discussion in https://github.com/sass/sass/issues/3171 about that, as I find that this makes modules harder to use (the whole point of modules was to make it easier to reason about files locally, but we can still be impacted by assignments done anywhere else).

I think it would make sense to have a linter rule forbidding assignment to namespaced variables (which does not cover the case of `@use "some-module" as *` but covers all other cases):

```scss
@use "some-module" as identifier;

identifer.$var: "should be rejected";
```

The idea would be to forbid variable assignments of the form `<identifier>.$<identifier>`

stof:
Closes #556

### DIFF
--- a/src/rules/dollar-variable-no-namespaced-assignment/index.js
+++ b/src/rules/dollar-variable-no-namespaced-assignment/index.js
@@ -1,0 +1,31 @@
+import { utils } from "stylelint";
+import { namespace } from "../../utils";
+
+export const ruleName = namespace("dollar-variable-no-namespaced-assignment");
+
+export const messages = utils.ruleMessages(ruleName, {
+  rejected: "Unexpected assignment to a namespaced $ variable"
+});
+
+export default function(actual) {
+  return (root, result) => {
+    const validOptions = utils.validateOptions(result, ruleName, { actual });
+
+    if (!validOptions) {
+      return;
+    }
+
+    root.walkDecls(decl => {
+      if (!/^[^$.]+\.\$./.test(decl.prop)) {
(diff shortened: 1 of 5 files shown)


src/rules/dollar-variable-no-namespaced-assignment/index.js
```diff
@@ -0,0 +1,31 @@
+import { utils } from "stylelint";
+import { namespace } from "../../utils";
+
+export const ruleName = namespace("dollar-variable-no-namespaced-assignment");
+
+export const messages = utils.ruleMessages(ruleName, {
+  expected: "Expected $ variable name to match specified pattern"
```
kristerkari, line 7: Same thing in this rule as in the other one:

I went and checked from the stylelint repo, and most of the `-no-` rules seem to use the `rejected: "Unexpected ..."` message format, should this rule also use it?

stof: I fixed it (and the whole message was wrong anyway due to a bad copy-paste)

kristerkari (approved):
Looks good! This can probably go to the next release (v4.1.0)

kristerkari:
This was released in 4.1.0:

Thank you @stof !

PR #580, opened 2021-12-05
Consider the use of sass modules in at-each-key-value-single-line rule

jhae-de:
The `at-each-key-value-single-line` rule does not report an error when set to `true` and using sass modules.

This should fail but it doesn't:
```scss
@use 'sass:map';

$font-weights: ("regular": 400, "medium": 500, "bold": 700);
@each $key in map.keys($font-weights) {
  $value: map.get($font-weights, $key);
}
```

This pull request changes the rule to take the use of sass modules into account.

### DIFF
--- a/src/rules/at-each-key-value-single-line/index.js
+++ b/src/rules/at-each-key-value-single-line/index.js
@@ -1,5 +1,5 @@
 import { utils } from "stylelint";
-import { namespace, moduleNamespace } from "../../utils";
+import { moduleNamespace, namespace } from "../../utils";
 
 export const ruleName = namespace("at-each-key-value-single-line");
 
@@ -77,13 +77,17 @@ function separateEachParams(paramString) {
 }
 
 function didCallMapKeys(mapDecl, mapNamespace) {
-  const pattern = getNamespacedPattern("keys\\(.*\\)", mapNamespace);
+  const pattern = getNamespacedPattern("keys\\(.*\\)", mapNamespace, mapDecl);
 
   return new RegExp(pattern).test(mapDecl);
 }
 
 function didCallMapGet(mapDecl, mapNamespace) {
-  const pattern = getNamespacedPattern("get\\((.*),(.*)\\)", mapNamespace);
+  const pattern = getNamespacedPattern(
+    "get\\((.*),(.*)\\)",
+    mapNamespace,
+    mapDecl
+  );
 
   return new RegExp(pattern).test(mapDecl);
 }
@@ -91,7 +95,11 @@ function didCallMapGet(mapDecl, mapNamespace) {
 // Fetch the name of the map from a map-keys() or map.keys() call.
 function mapName(mapDecl, mapNamespace) {
   if (didCallMapKeys(mapDecl, mapNamespace)) {
-    const pattern = getNamespacedPattern("keys\\((.*)\\)", mapNamespace);
+    const pattern = getNamespacedPattern(
+      "keys\\((.*)\\)",
+      mapNamespace,
+      mapDecl
+    );
 
     return mapDecl.match(new RegExp(pattern))[1];
   }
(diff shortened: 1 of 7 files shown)


kristerkari:
Thanks! I'll try to review asap

stof:
This should support cases where the module is imported with a different namespace than the default `map` to be complete

jhae-de:
Namespace support added. Thanks for your advice.

kristerkari:
@jhae-de great job with adding so many test cases! I still need to go through the changes more in detail before I can approve.

stof (changes requested):

src/utils/moduleNamespace.js
```diff
@@ -32,7 +32,10 @@ export function moduleNamespace(root, module) {
       return;
     }
 
-    moduleNamespace = parts[1].trim();
+    moduleNamespace = parts[1]
+      .trim()
+      .split(" ")[0]
```
stof, line 37: be careful. `as*with` could be written without spaces (as `*` is not a valid identifier character, spaces are optional)

jhae-de:
Thank you for your review @stof. I added more test cases and hopefully solved all problems.

src/rules/at-each-key-value-single-line/index.js
```diff
@@ -102,19 +110,26 @@ function mapName(mapDecl, mapNamespace) {
...
   }
 
-  return `${mapNamespace}[-.]${pattern}`;
+  if (namespace === "*") {
+    return pattern;
+  }
+
+  return new RegExp(`${namespace}.${pattern}`).test(decl)
+    ? `${namespace}.${pattern}`
+    : `map-${pattern}`;
```
stof, line 134: I suggest returning a pattern `(${namespace}\.|map-)${pattern}`, which avoids having to test `decl` (and it could work with the default namespace too, by making it be treated as `map`)

src/rules/at-each-key-value-single-line/index.js
```diff
@@ -102,19 +110,26 @@ function mapName(mapDecl, mapNamespace) {
...
+    return `map[-.]${pattern}`;
   }
 
-  return `${mapNamespace}[-.]${pattern}`;
+  if (namespace === "*") {
+    return pattern;
+  }
+
+  return new RegExp(`${namespace}.${pattern}`).test(decl)
+    ? `${namespace}.${pattern}`
```
stof, line 133: this misses escaping the `.`, which would make it match things like `miget` for the `m` namespace

src/utils/moduleNamespace.js
jhae-de: Sure. And since we're already talking about it, there are other cases such as:

```scss
// With line breaks
@use 'sass:map'
  as
  *
  with ();
```
or:
```scss
// Using 'with' as namespace
@use 'sass:map'
  as
  with
  with ();
```

That seems to escalate a bit,
but I will have a look at it.

src/rules/at-each-key-value-single-line/index.js
jhae-de: That is of course much better. I'm not happy with the current implementation anyway.

stof: Speeking of that, I think `moduleNamespace()` might actually be nicer if it was returning the default namespace instead of `null` (i.e. the basename of the module URL), and returning `null` instead of `*` when using a module without namespace. This would have 2 benefits:

- the rule code would not need to have a special case for default namespaces (as that would work the same than for an explicit namespace)
- it would be consistent with the way Sass itself treats the namespace for those `@use` rules, making it less confusing for Sass contributors to contribute here with those helpers (for people that don't contribute to Sass itself, they would not know about the inconsistency and so would not care)

jhae-de: Sounds good, let's collect some constellations:

```scss
@use 'sass:color'; // namespace 'color'
@use 'my-module'; // namespace 'my-module'
@use 'my-module.scss'; // namespace 'my-module'
@use 'path/to/my-module'; // namespace 'my-module'
@use 'path/to/my-module.scss'; // namespace 'my-module'
```

Am I missing other possibilities?

stof: I think this is OK (if you account for the `.sass` extension too)

kristerkari:
I see that @stof has done a great job with the review here, so I trust that he can also say if he thinks that this is good to be merged now :) At least there are a lot of test cases!

kristerkari:
@stof Could you please re-review this (as you have already reviewed the PR) to see if everything is ok now?

I could release a new minor version with this and the other PRs.

stof (approved):

kristerkari:
This was released in 4.1.0:

Thank you @jhae-de and @stof !

PR #591, opened 2022-02-14
Add `scss/function-no-unknown` rule

### LINKED ISSUE #581: Feature request: `scss/function-no-unknown`
### Problem

Sass doesn't catch typos in function names. The docs [recommend](https://sass-lang.com/documentation/at-rules/function#plain-css-functions) using a linter to catch these issues:

>Because any unknown function will be compiled to CSS, it’s easy to miss when you typo a function name. Consider running a [CSS linter](https://stylelint.io/) on your stylesheet’s output to be notified when this happens!

I'm unable to find any rules in stylelint-scss that catch mistakes like these, however.

### Proposed solution

It would be awesome to have a new rule to catch mistakes like these. It would work like `scss/at-rule-no-unknown`, but for functions. Let's call it `scss/function-no-unknown`. 

### Example

Here's an example of a mistake. I've typed `invart` instead of `invert`:
```scss
@function invert($color, $amount: 100%) {
  $inverse: change-color($color, $hue: hue($color) + 180);
  @return mix($inverse, $color, $amount);
}

.button {
  background-color: invart(#f00)
}
```

Sass assumes that `invart` must be a CSS function and writes it into the compiled CSS:

```css
.button {
  background-color: invart(#f00)
}
```

Our proposed rule would realize that there's no Sass function called `invart` and also no CSS function called `invart`. It would throw an error, alerting us to the fact that we've made a typo.

Thank you so much for your time! :)

### LINKED ISSUE #589: New stylelint `function-no-unknown` rule not working well with SCSS
A new rule `function-no-unknown` was recently added to `stylelint`, merged by @jeddy3 for stylelint/stylelint#5853. This rule considers functions defined in the CSS Specifications to be 'known', and ignores double-dashed custom functions, _e.g._ `--custom-function()`. This rule seems good and useful, and the recommendation to use double-dashed names so as to avoid clashes with current or future standard CSS names would seem applicable to SASS/SCSS functions. However, I have encountered some problems when applying it in conjunction with `stylelint-config-standard-scss` and linting SCSS files. I don't know whether the issue is one for `stylelint-config-standard-scss` to address or whether it needs addressing in base `stylelint`, or a combination of both, so I have opened this issue here to track it.

### Problem 1

Standard functions defined by SASS are not 'known' by this rule, even if `stylelint-config-standard-scss` is being used. This causes many errors to be produced such as:
- Unexpected unknown function "if"
- Unexpected unknown function "map-get"
- _etc_

This is easily worked around by adding these built-in function names to the rule using `ignoreFunctions`, but that is a manual and error-prone process. SASS built-in functions should be automatically included as 'known' when SASS/SCSS files are being linted.

### Problem 2

The rule does not play well with interpolation, as it doesn't recognise the syntax of `#{...}`. For example, a line such as
```scss
      transform: translateX(calc(-1 * #{map-get($side-panel-sizes, md)}));
```
produces an error such as
- Unexpected unknown function "#{map-get"

This error is poorly formed, and note that this also means that even if the function which appears at the start of the `#{...}` interpolation is a 'known' function or is double-dashed it will still be reported as an error. For example:
- Unexpected unknown function "#{--custom-function".

Again, this is easily worked around by including the names of affected functions including the wrongly-included `#{` into `ignoreFunctions`, but this is rather fiddly and far from ideal. The rule should understand the parsed syntax of SASS interpolation and correctly identify function names within this.

ybiquitous:
- Closes #581
- Closes #589

### DIFF
--- a/src/rules/function-no-unknown/README.md
+++ b/src/rules/function-no-unknown/README.md
@@ -1,0 +1,64 @@
+# function-no-unknown
+
+Disallow unknown functions. Should be used **instead of** Stylelint's [function-no-unknown](https://stylelint.io/user-guide/rules/list/at-rule-no-unknown).
+
+```css
+a { color: unknown(1); }
+/**        ↑
+ * Functions like this */
+```
+
+This rule is basically a wrapper around the mentioned core rule. You must disable Stylelint's core rule to make this rule work:
+
+```json
+{
+  "rules": {
+    "function-no-unknown": null,
+    "scss/function-no-unknown": true
+  }
+}
+```
+
+## Options
+
+### `true`
+
+The following patterns are considered warnings:
+
+```css
+a { color: unknown(1); }
+```
+
+The following patterns are *not* considered warnings:
+
+```css
+a { color: hsl(240 100% 50%); }
+```
+
+```css
+a { color: if(true, green, red); }
+```
+
+## Optional secondary options
+
+### `ignoreFunctions: ["/regex/", /regex/, "non-regex"]`
+
+Given:
+
+```json
+["/^my-/i", "foo"]
+```
+
+The following patterns are *not* considered warnings:
+
+```css
+a { color: my-func(1); }
+```
+
+```css
+a { color: MY-FUNC(1); }
+```
+
+```css
+a { color: foo(1); }
+```
--- a/src/rules/function-no-unknown/__tests__/index.js
+++ b/src/rules/function-no-unknown/__tests__/index.js
@@ -1,0 +1,73 @@
+import { ruleName, messages } from "..";
+
+testRule({
+  ruleName,
+  config: [true],
+  customSyntax: "postcss-scss",
+
+  accept: [
+    {
+      code: "a { color: hsl(240 100% 50%); }"
+    },
+    {
+      code: "a { color: if(true, green, red); }"
+    },
+    {
+      code: "a { color: adjust-color(#6b717f, $red: 15); }"
+    },
+    {
+      code: "a { color: color.adjust(#6b717f, $red: 15); }"
+    }
+  ],
+
+  reject: [
+    {
+      code: "a { color: unknown(1); }",
+      message: messages.rejected("unknown"),
+      line: 1,
+      column: 12
+    }
+  ]
+});
+
+testRule({
+  ruleName,
+  config: [true, { ignoreFunctions: ["/^my-/i", /foo$/, "bar"] }],
+  customSyntax: "postcss-scss",
+
+  accept: [
+    {
+      code: "a { color: my-func(1); }"
+    },
+    {
+      code: "a { color: MY-FUNC(1); }"
+    },
+    {
+      code: "a { color: func-foo(1); }"
+    },
+    {
+      code: "a { color: bar(1); }"
+    }
+  ],
+
+  reject: [
+    {
+      code: "a { color: my(1); }",
+      message: messages.rejected("my"),
+      line: 1,
+      column: 12
+    },
+    {
+      code: "a { color: foo-func(1); }",
+      message: messages.rejected("foo-func"),
+      line: 1,
+      column: 12
+    },
+    {
+      code: "a { color: barrr(1); }",
+      message: messages.rejected("barrr"),
+      line: 1,
+      column: 12
+    }
+  ]
+});
--- a/src/rules/function-no-unknown/__tests__/index.js
+++ b/src/rules/function-no-unknown/__tests__/index.js
@@ -26,6 +26,12 @@ testRule({
       message: messages.rejected("unknown"),
       line: 1,
       column: 12
+    },
+    {
+      code: "a { color: color.unknown(#6b717f, $red: 15); }",
+      message: messages.rejected("color.unknown"),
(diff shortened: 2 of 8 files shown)


src/utils/functions.js
```diff
@@ -0,0 +1,42 @@
...
+  "scale-color",
+  "transparentize"
+]);
+
+// TODO: Add other functions...
+
+export const ALL_FUNCTIONS = Object.freeze([
+  ...GLOBAL_FUNCTIONS,
+  ...COLOR_FUNCTIONS
+]);
```
ybiquitous, line 42: [ask] Is this a smart way to get all Sass functions?

src/rules/function-no-unknown/__tests__/index.js
```diff
@@ -0,0 +1,73 @@
...
+      code: "a { color: hsl(240 100% 50%); }"
+    },
+    {
+      code: "a { color: if(true, green, red); }"
+    },
+    {
+      code: "a { color: adjust-color(#6b717f, $red: 15); }"
+    },
+    {
+      code: "a { color: color.adjust(#6b717f, $red: 15); }"
```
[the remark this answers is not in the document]
kristerkari: One option is simply not to support `@use "sass:color" as c` if it gets too difficult to implement.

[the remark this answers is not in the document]
ybiquitous: Indeed, as the first implementation, not supporting `as` sounds good. 👍🏼

ybiquitous:
I've bumped `stylelint` to [14.5.1](https://github.com/stylelint/stylelint/releases/tag/14.5.1). Now the CI is all green!

kristerkari:
@ybiquitous Does this rule now depend on stylelint version 14.5.1 or newer or is there backwards compatibility?

ybiquitous:
@kristerkari 

> Does this rule now depend on stylelint version 14.5.1 or newer

Yes. The new rule requires `stylelint@14.5.1` or newer.

> is there backwards compatibility?

Yes. Unless users don't enable the new rule, their stylelint analysis should not change.

src/rules/function-no-unknown/README.md
```diff
@@ -0,0 +1,64 @@
+# function-no-unknown
+
+Disallow unknown functions. Should be used **instead of** Stylelint's [function-no-unknown](https://stylelint.io/user-guide/rules/list/at-rule-no-unknown).
```
kristerkari, line 3: @ybiquitous if this is meant to be used instead of `function-no-unknown`, then shouldn't this rule be checking against normal CSS functions too?

ybiquitous: Actually, the current implementation in this pull request does it:

```js
const ruleToCheckAgainst = "function-no-unknown";

// ...

utils.checkAgainstRule({ ruleName: ruleToCheckAgainst, ... });
```

1. Check against normal CSS functions via `utils.checkAgainstRule()`
2. Check against SCSS functions via `if (!ignoreFunctionsAsSet.has(funcName)) { ... }`

kristerkari: okay, I see.

In that case could you also add some kind of unit test to verify that there is a warning for CSS functions? Just to make sure that the rule is working correctly.

ybiquitous: I've pushed 42889c2. Are the following cases enough?

- `unknown()` case for CSS function
- `color.unknown()` case for SCSS function

```js
  reject: [
    {
      code: "a { color: unknown(1); }",
      message: messages.rejected("unknown"),
      line: 1,
      column: 12
    },
    {
      code: "a { color: color.unknown(#6b717f, $red: 15); }",
      message: messages.rejected("color.unknown"),
      line: 1,
      column: 12
    }
  ]
```

kristerkari: Oh yeah I wasn't thinking properly when I wrote my comment, sorry. 

What I was thinking about was that you could add a test that some non-SCSS function from this list: https://github.com/niksy/css-functions-list/blob/master/index.json passes and does not give a warning. 

But anyway those reject tests are also good to have 👍

ybiquitous: I understand your intention well. 👍🏼 

I've pushed 5dde89e to add more *accept* cases.

kristerkari: Very nice, thanks!

kristerkari (approved):
Looks good now! 👍

kristerkari:
Thanks @ybiquitous ! Released now in https://github.com/stylelint-scss/stylelint-scss/releases/tag/v4.2.0

ybiquitous:
@kristerkari Thank you! 😊

chalkygames123:
@kristerkari Hi, any plan to include this rule to `stylelint-config-recommended-scss`?

ybiquitous:
@chalkygames123 Here is a reason not to include this rule to `stylelint-config-recommended-scss`:
https://github.com/stylelint-scss/stylelint-config-recommended-scss/issues/92#issuecomment-1068882019

chalkygames123:
I see. Thanks for letting me know!

PR #602, opened 2022-05-09
Add `meta` property for all rules

ybiquitous:
Each rule function should have properties like `ruleName`.
See https://stylelint.io/developer-guide/plugins

For example, `stylelint -f verbose` command can show rules' links:

image

### DIFF
--- a/src/utils/ruleUrl.js
+++ b/src/utils/ruleUrl.js
@@ -1,0 +1,15 @@
+const revision = "master";
+
+/**
+ * Get the specified rule's URL.
+ *
+ * @param {string} ruleName
+ * @return {string}
+ */
+export function ruleUrl(ruleName) {
+  let name = ruleName;
+  if (name.includes("/")) {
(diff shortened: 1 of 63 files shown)


src/utils/ruleUrl.js
```diff
@@ -0,0 +1,15 @@
+const revision = "master";
```
ybiquitous, line 1: [note] If using a tag (e.g. `v4.2.0`), the implementation would be more complex. I think the HEAD version (`master`) is enough.

kristerkari:
Thanks! I'll try to get this reviewed ASAP.

src/utils/ruleUrl.js
kristerkari: Using `master` sounds good enough.

kristerkari (approved):
Thanks! Looks good 👍

ybiquitous:
@kristerkari Thank you for the review and merge! 😄

Issue #606, opened 2022-05-16
`scss/comment-no-empty` false positive on paragraph breaks in multi-line double-slash comments


dbjorge:
Comments like this:

```scss
// First paragraph of a long comment
//
// Second paragraph
.some-rule { /* ... */ }
```

...trigger a false positive from `scss/comment-no-empty` on the blank line within the comment acting as a paragraph break. Here is an example `accept` test case that can be added to `/src/rules/comment-no-empty/__tests__/index.js` which demonstrates the issue:

```js
    {
      code: "// comment\n//\n// comment",
      description: "Multiline double slash comment with paragraph break"
    },
```

```
  ● scss/comment-no-empty › accept › [ true ] › '// comment\n//\n// comment' › Multiline double slash comment with paragraph break

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 9

    - Array []
    + Array [
    +   Object {
    +     "column": 1,
    +     "line": 2,
    +     "rule": "scss/comment-no-empty",
    +     "severity": "error",
    +     "text": "Unexpected empty comment (scss/comment-no-empty)",
    +   },
    + ]

      at Object.<anonymous> (node_modules/jest-preset-stylelint/getTestRule.js:61:41)
```

PR #607, opened 2022-05-16
Make `scss/comment-no-empty` treat multiline double-slash comments as blocks

dbjorge:
Proposed fix for #606

I've never contributed to a stylelint rule before, so apologies if I've made any structural mistakes here. In particular, I felt a bit uncertain about adding state that persists between different calls to the `walkComments` callback, since I didn't see other examples of rules doing that, but it seemed like the most straightforward way to solve the problem.

The logic for determining whether a given comment is double-slash vs not and standalone vs not is taken from the examples on L44-51 of `src/rules/double-slash-comment-empty-line-before/index.js`.

### DIFF
--- a/src/rules/comment-no-empty/__tests__/index.js
+++ b/src/rules/comment-no-empty/__tests__/index.js
@@ -17,6 +17,33 @@ testRule({
     {
       code: "/* comment */",
       description: "Single line block comment"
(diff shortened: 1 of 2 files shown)


AlexWayfer:
Faced such problem.

PR #608, opened 2022-05-19
Add `at-import-partial-extension` autofix only when `"never"`

ybiquitous:
This PR adds the *autofix* support to the `scss/at-import-partial-extension` rule only when the `"never"` option is given.

I believe we cannot support autofix when `"always"` is given because it's impossible to determine file extensions to be added.


kristerkari:
Thanks! More stuff for my review queue :D

kristerkari (approved):
Looks good 👍

PR #609, opened 2022-05-19
Fix `--fix` option link on each rule doc

ybiquitous:
For example, the doc of a core rule `alpha-value-notation` is helpful.
See https://stylelint.io/user-guide/rules/list/alpha-value-notation


kristerkari:
Thanks! Looks good 👍

PR #612, opened 2022-06-29
Bug fix: Make filenames with dots work with at-import-partial-extension rule

kersh:
### Bug
- Related issue: https://github.com/stylelint-scss/stylelint-scss/issues/587#issuecomment-1025651087
- Fixing an issue when using `@import` with scss files that contain `.` symbol.
image

**Before (didn't work)**
```scss
@import "colors.variables"; // <--- Stylelint: Unexpected extension ".utility" in @import (scss/at-import-partial-extension)
@import "defaultOutlineStyle.mixin"; // <--- Stylelint: Unexpected extension ".utility" in @import (scss/at-import-partial-extension)

// ...
```

**After (works now)**
```scss
@import "colors.variables"; // OK
@import "defaultOutlineStyle.mixin"; // OK

// ...
```

### DIFF
--- a/src/rules/at-import-partial-extension/index.js
+++ b/src/rules/at-import-partial-extension/index.js
@@ -75,7 +75,8 @@ export default function rule(expectation, _, context) {
           return;
         }
 
-        if (extension && expectation === "never") {
+        const isScssPartial = extension === "scss";
+        if (extension && isScssPartial && expectation === "never") {
           if (context.fix) {
             const extPattern = new RegExp(`\\.${extension}(['" ]*)`, "g");
             decl.params = decl.params.replace(extPattern, "$1");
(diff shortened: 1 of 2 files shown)


src/rules/at-import-partial-extension/index.js
```diff
@@ -75,7 +75,8 @@ export default function rule(expectation, _, context) {
           return;
         }
 
-        if (extension && expectation === "never") {
+        const isScssPartial = extension === "scss";
+        if (extension && isScssPartial && expectation === "never") {
```
kersh, line 79: So `extension` contains after dot value (in my case it is `.variables` or `.mixin`).
I assume we only care about checking if there is a `.scss` file partial since we work with Sass. Other `extension` values can be left in the code and linter shouldn't highlight them.

kristerkari: 👍

kristerkari (approved):
Looks good 👍

PR #613, opened 2022-07-05
New Rule: `at-extend-disallow`

DerekTBrown:
I would like to introduce a new rule,`at-extend-disallow`, which would allow developers to lint against the use of `@extend` altogether. We see several issues with `@extend` in practice:

1. `@extend` [obfuscates the meaning](https://pressupinc.com/blog/2014/11/dont-overextend-yourself-in-sass/) of a particular rule.
2. `@extend` leads to [specificity and inheritance problems](https://webinista.com/updates/dont-use-extend-sass/), which are hard to troubleshoot.
3. `@extend` behaves differently across [versions of Sass](https://sass-lang.com/documentation/breaking-changes/extend-compound).

Instead, we encourage developers to use Mixins, which offer similar functionality without some of these drawbacks. 

It seems that `sass-lint` currently offers a similar [`no-extends`](https://github.com/sasstools/sass-lint/blob/develop/docs/rules/no-extends.md) rule, so I would like to propose we support one in `stylelint-scss` as well.

### DIFF
--- a/src/rules/at-extend-disallow/README.md
+++ b/src/rules/at-extend-disallow/README.md
@@ -1,0 +1,3 @@
+# at-extend-disallow
+
+Disallow `@extend` declarations of any type.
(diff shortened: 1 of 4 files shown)


DerekTBrown:
Ahh, I just found `at-rule-disallowed-list`. I think this will work for us.

PR #639, opened 2022-07-19
Fix at-rule pattern warnings to range only one line

nlydv:
Hi, first contrib here.

I noticed the `at-function-pattern` and `at-mixin-pattern` rules spit out warnings with a range over the *entire* at-rule block as the "problem area", which of course doesn't make sense when it's just the name that's the issue.

Not at all familiar with the codebase and I just wanted to slap a quick fix on it if I could, best I was able to do was limiting it to one line only.

Before:
before

After:
after

### DIFF
--- a/src/rules/at-function-pattern/__tests__/index.js
+++ b/src/rules/at-function-pattern/__tests__/index.js
@@ -77,6 +77,9 @@ testRule({
       }
     `,
       line: 2,
+      column: 7,
+      endLine: 3,
+      endColumn: 0,
       message: messages.expected,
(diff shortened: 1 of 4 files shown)


kristerkari:
@nlydv Thanks a lot! Could you please also update the unit tests to verify that the change is working? 

I think that you should be able to assert the `endLine` and `endColumn` like here for example:
https://github.com/stylelint/stylelint/blob/dab8df2b3bf58e545f683a0f49d821abc089cde3/lib/rules/comment-word-disallowed-list/__tests__/index.js#L22-L25

PR #643, opened 2022-08-01
bugfix(at-import-partial-extension): fixer incorrectly replaces filename

chimurai:
Hi, my first contrib here. 👋

While fixing a large codebase I noticed the `at-import-partial-extension` fixer incorrectly fixed certain files when they have `.scss` in their filename. (example: `component.scss-theme.scss`)

---

**currently**:
`at-import-partial-extension`'s fixer incorrectly replaced all `.scss` occurences in a filename.

```diff
- @import "component.scss-theme.scss";
+ @import "component-theme";
```

**expected**:
`at-import-partial-extension`'s fixer should only replace the `.scss` extension (not when `.scss` is used in the filename itself)

```diff
- @import "component.scss-theme.scss";
+ @import "component.scss-theme";
```

### DIFF
--- a/src/rules/at-import-partial-extension/__tests__/index.js
+++ b/src/rules/at-import-partial-extension/__tests__/index.js
@@ -406,6 +406,18 @@ testRule({
       column: 33,
       message: messages.rejected("scss"),
       description: "Single file, has .scss extension and a dot in the name."
+    },
+    {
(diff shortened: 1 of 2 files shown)


kristerkari:
Thanks @chimurai ! Looks really good!

PR #644, opened 2022-08-01
ci(husky): fix husky v8

chimurai:
Bumping husky from v4 to v8 broke husky.

Introduced in change: https://github.com/stylelint-scss/stylelint-scss/pull/627

Followed migration guide: https://github.com/typicode/husky-4-to-8

Side note: Prettier2 upgrade (https://github.com/stylelint-scss/stylelint-scss/pull/625) will format files slightly differently.


kristerkari:
Thanks @chimurai !

Issue #652, opened 2022-08-18
Support end positions for warnings


ybiquitous:
[Stylelint 14.7.0](https://github.com/stylelint/stylelint/releases/tag/14.7.0) has supported end positions for warnings, and the official VSCode extension [prepares the end position support](https://github.com/stylelint/vscode-stylelint/pull/358).

So, I suggest `stylelint-scss` also supports end positions.

Currently, `stylelint-scss` warns for the whole node, but it seems not good DX:

<img width="1031" alt="image" src="https://user-images.githubusercontent.com/473530/185413438-50ea0ad6-0d5f-4d8b-91b7-fb603b37b819.png">

<details>
<summary>Reproduction Details</summary>

`package.json`:

```json
{
  "dependencies": {
    "postcss-scss": "4.0.4",
    "stylelint": "14.10.0",
    "stylelint-codeframe-formatter": "1.1.0",
    "stylelint-scss": "4.3.0"
  },
  "scripts": {
    "test": "stylelint \"*.scss\" --custom-formatter stylelint-codeframe-formatter"
  }
}
```

`.stylelintrc.js`:

```js
module.exports = {
  plugins: ["stylelint-scss"],
  customSyntax: "postcss-scss",
  rules: {
    "scss/at-rule-no-unknown": true,
    "scss/at-if-no-null": true,
  },
};
```

`a.scss`:

```scss
@foo {
  // ...
}

a {
  @if $x == null {}
}
```

</details>

---

This will change many files (maybe all rules), but I can help from [my experience in the stylelint/stylelint repo](https://github.com/stylelint/stylelint/issues/5694).

kristerkari:
Thanks! This would definitely be a good improvement! :)

ybiquitous:
@kristerkari Thanks for the response. It seems this task needs many PRs, so I create the following list to make the progress clear:

- [ ] [at-each-key-value-single-line](https://github.com/stylelint-scss/stylelint-scss/tree/master/src/rules/at-each-key-value-single-line)
- [ ] [at-else-closing-brace-newline-after](https://github.com/stylelint-scss/stylelint-scss/tree/master/src/rules/at-else-closing-brace-newline-after)
- [ ] [at-else-closing-brace-space-after](https://github.com/stylelint-scss/stylelint-scss/tree/master/src/rules/at-else-closing-brace-space-after)
[...]

PR #655, opened 2022-08-23
Fix `at-rule-no-unknown` end positions

ybiquitous:
Part of #652

This change also adds a test helper function to make testing multiline strings easier.

### DIFF
--- a/src/rules/at-rule-no-unknown/__tests__/index.js
+++ b/src/rules/at-rule-no-unknown/__tests__/index.js
@@ -7,115 +7,111 @@ testRule({
 
   accept: [
     {
-      code: "@charset 'UTF-8';"
+      code: "@charset 'UTF-8';",
     },
     {
-      code: "@CHARSET 'UTF-8';"
+      code: "@CHARSET 'UTF-8';",
     },
     {
-      code: "@charset 'iso-8859-15'"
+      code: "@charset 'iso-8859-15'",
     },
     {
-      code: '@import url("fineprint.css") print;'
+      code: '@import url("fineprint.css") print;',
     },
     {
-      code: "@import 'custom.css'"
+      code: "@import 'custom.css'",
     },
     {
-      code: "@import url('landscape.css') screen and (orientation:landscape);"
+      code: "@import url('landscape.css') screen and (orientation:landscape);",
     },
     {
-      code: "@namespace url(http://www.w3.org/1999/xhtml);"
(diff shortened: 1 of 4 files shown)


src/rules/at-rule-no-unknown/__tests__/index.js
```diff
@@ -7,195 +7,191 @@ testRule({
...
-      @funciton floo ($n) {}
-    `,
-      line: 2,
+      code: dedent`
+        @funciton floo ($n) {}
+      `,
+      line: 1,
+      column: 1,
+      endLine: 1,
+      endColumn: 10,
```
ybiquitous, line 112: [note] It seems to me that one case change is enough to test end positions.

src/rules/at-rule-no-unknown/index.js
```diff
@@ -83,8 +83,8 @@ export default function rule(primaryOption, secondaryOptions) {
             ruleName,
             result,
             node: warning.node,
-            line: warning.line,
-            column: warning.column
+            start: { line: warning.line, column: warning.column },
+            end: { line: warning.endLine, column: warning.endColumn },
```
kristerkari, line 87: @ybiquitous is this a backwards compatible change?

ybiquitous: Good point. I tried testing the minimum version of `peerDependencies.stylelint` locally:

https://github.com/stylelint-scss/stylelint-scss/blob/efd565d2e0f1bd1c70c4242a94be200795dd5e51/package.json#L40-L42

```console
$ npm i stylelint@14.5.1
...

$ npx jest at-rule-no-unknown --no-verbose
 FAIL  src/rules/at-rule-no-unknown/__tests__/index.js
  ● scss/at-rule-no-unknown › reject › [ true ] › '@funciton floo ($n) {}\n' › no description

    expect(received).toMatchObject(expected)

    - Expected  - 2
    + Received  + 0

      Object {
        "column": 1,
    -   "endColumn": 10,
    -   "endLine": 1,
        "line": 1,
        "text": "Unexpected unknown at-rule \"@funciton\" (scss/at-rule-no-unknown)",
      }

      at Object.<anonymous> (node_modules/jest-preset-stylelint/getTestRule.js:88:33)
```

It seems that the backward compatibility keeps. 👍🏼 

If possible, I think it's better to update `peerDependencies.stylelint` to `^14.7.0`, but not required.

ybiquitous:
Don't mind. Thanks for the merge!

PR #689, opened 2022-10-22
Fix `at-each-key-value-single-line` end positions

ybiquitous:
Part of #652

See the following example to be reported:

```scss
@each $key in map-keys($font-weights) {
/*    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 *    Use @each $key, $value in $map syntax instead of $value: map-get($map, $key)
 */
  $value: map-get($font-weights, $key);
}
```


kristerkari (approved):
Thank you 👍

PR #698, opened 2022-11-02
Add failing test for function-no-unknown with negative values

amk221:
stylelint-scss seems to be detecting `-` as a function name.

<hr>

Please note that my PR automatically swapped double quotes for single quotes with a pre-commit hook. I believe it ran prettier, but since there is no prettier config in this repo, it used my default config. Let me know what to do and I'll fix it up.


XhmikosR:
@amk221 feel free to make a new PR from master without the redundant changes.

Issue #786, opened 2023-03-21
Interpolation in calc() function variable


danny007in:
Rule Example

Allow
```css
 border-top: calc(#{$table-border-width} * 2) 
```
Disallow
```css
 border-top: calc($table-border-width * 2) 
```

To solve https://github.com/twbs/bootstrap/issues/38278

kristerkari:
@danny007in Maybe this could be a more generic rule, like disallowing interpolation in function call arguments?

danny007in:
PR Created in https://github.com/stylelint-scss/stylelint-scss/pull/787

kristerkari:
@danny007in Thanks! I just realized that you want the rule to do the opposite of what I was saying in my previous comment.

 I'm still however thinking if there is a need for this kind of rule in `stylelint-scss`. It seems very specific to the problem with `calc`, and I'm wondering if this is a common problem at all? 

The other option would be just to add a custom rule to the project in question (bootstrap) to catch this specific calc related problem.

danny007in:
@kristerkari to solve https://github.com/twbs/bootstrap/issues/38278

stof:
The rule requested here is related to the unmaintained node-sass only. When using a modern Sass compiler supporting first-class calc, the best practice is precisely the opposite: not using interpolation in `calc()` but using SassScript expressions directly.

PR #787, opened 2023-03-22
New Feature calc-variable-interpolation

danny007in:


kristerkari:
As per discussion in #786, let's close this one and instead improve the `dollar-variable-no-missing-interpolation` to support calc.

PR #794, opened 2023-03-24
Move things outside of package.json

XhmikosR:
@kristerkari I know you might not prefer this solution but our package.json keeps growing and it's harder to find things.

### DIFF
--- a/.babelrc.json
+++ b/.babelrc.json
@@ -1,0 +1,3 @@
+{
+  "presets": ["@babel/env"]
+}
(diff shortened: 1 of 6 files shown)


kristerkari (approved):
Actually I don't mind having the configs in separate files. I have just usually had them in `package.json` to avoid at lot of extra files in the project root.

I think that this change is totally fine 👍

XhmikosR:
BTW @kristerkari feel free to backport CI and stuff from this repo to the other repos if you want. Everything is more tweaked in this repo. :)

kristerkari:
@XhmikosR sure, it's just usually a question of having time to do it as I think that I have ~10 stylelint plugins and configs :)

PR #795, opened 2023-03-25
Use optional chaining instead of dlv

XhmikosR:
I'm not 100% what can be `null`/`undefined` so I went with optional chaining in all paths.

If we know for some might not be nullish, we can drop some instances


kristerkari:
> while you review this (happy to skip it for the next release BTW) could you enable the repository option for the branches to be up to date and delete `HEAD` branches?

Sure, I'll have a look at those options.

XhmikosR:
Is node ever going to be null/undefined? If not, I think we can drop the first optional chaining after `node`.

XhmikosR:
Let's leave it as is then. Feel free to merge at your leisure :)

PR #802, opened 2023-04-06
Remove babel

XhmikosR:
@kristerkari here is the old `rm-babel` branch rebased and updated.

I don't quite get the failures right now, another pair of eyes would help for sure.

**I suggest that you cut a minor release from master and then we try to move with this PR so that we can cut a new major version.**

Feel free to push to the branch if you have any fixes :)

EDIT: I think I know why the failures happen. I must have messed up somewhere the export in rules/index.js and/or utils/index.js.

Let me know if you have any ideas!


XhmikosR:
Also, CC @ybiquitous in case you can help me out here :)

I must be missing something because my previous branch some years ago worked. https://github.com/stylelint-scss/stylelint-scss/tree/rm-babel

But there have been so many changes since then that I only keep the branch as reference until I'm done with the new one.

src/utils/index.js
```diff
@@ -1,35 +1,40 @@
...
+  optionsHaveException: require("./optionsHaveException"),
+  optionsHaveIgnored: require("./optionsHaveIgnored"),
+  parseFunctionArguments: require("./parseFunctionArguments"),
+  parseNestedPropRoot: require("./parseNestedPropRoot"),
+  parseSelector: require("./parseSelector"),
+  rawNodeString: require("./rawNodeString"),
+  removeEmptyLinesBefore: require("./removeEmptyLinesBefore"),
+  ruleUrl: require("./ruleUrl"),
+  findOperators: require("./sassValueParser"),
+  validateTypes: require("./validateTypes"),
```
ybiquitous, line 36: @XhmikosR This exporting `validateTypes` seems wrong. The original code is below:

https://github.com/stylelint-scss/stylelint-scss/blob/41b71f39a462f298ffdec394cbeb9bad14542f50/src/utils/index.js#L35

I don't know how to rewrite `export *` to CommonJS, but exporting each function seems to resolve test failures like this:

```javascript
const { isString, isRegExp } = require("./validateTypes");

const utils = {
  // ...
  isString,
  isRegExp
};
```

XhmikosR: Ah true, thanks!

I'm thinking maybe I'll just drop utils/index.js.

XhmikosR:
Getting closer, there are still some exports related problems with files that export more than one thing.

XhmikosR:
@kristerkari friendly ping!

>I suggest that you cut a minor release from master and then we try to move with this PR so that we can cut a new major version.

PS. I still think the option for branches to be up to date should be enabled in the repo. You can enable it in branch protection settings for master.

XhmikosR:
@kristerkari friendly ping, please check the OP and my comments above when you have a little time :)

kristerkari:
> PS. I still think the option for branches to be up to date should be enabled in the repo. You can enable it in branch protection settings for master.

Hmm.. I tried looking for that option, but could not find it.

I think that this PR is good to go, so I can have a look at first doing a small release with the master changes later today.

kristerkari (approved):
👍

PR #812, opened 2023-04-25
CI: add Node.js 20

XhmikosR:


kristerkari (approved):
Looks good 👍

XhmikosR:
Nice! Can you also cut a v5.0.0 release?

kristerkari:
yes, I can!

XhmikosR:
Thanks!

Just remember to use npm 8.x due to https://github.com/npm/cli/issues/6341. Not a huge deal if you don't, but it won't be how slim the tar would be if that bug didn't exist.

PR #815, opened 2023-05-02
Use import-lazy for rules

XhmikosR:
```
C:\Users\xmr\Desktop\stylelint-scss>git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.

C:\Users\xmr\Desktop\stylelint-scss>hyperfine "node load-time.js"
Benchmark 1: node load-time.js
  Time (mean ± σ):     174.5 ms ±   3.7 ms    [User: 0.8 ms, System: 2.5 ms]
  Range (min … max):   171.1 ms … 185.3 ms    17 runs

C:\Users\xmr\Desktop\stylelint-scss>git checkout -
Switched to branch 'xmr/import-lazy'

C:\Users\xmr\Desktop\stylelint-scss>hyperfine "node load-time.js"
Benchmark 1: node load-time.js
  Time (mean ± σ):     123.5 ms ±   5.9 ms    [User: 2.5 ms, System: 2.9 ms]
  Range (min … max):   116.4 ms … 140.6 ms    23 runs
```

load-time.js:

```js
"use strict";

console.log("Load time");
console.time("Module");

exports.m = require(".");

console.timeEnd("Module");
```

Approximately 28% faster load time on my machine.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "5.0.1",
       "license": "MIT",
       "dependencies": {
+        "import-lazy": "^4.0.0",
         "postcss-media-query-parser": "^0.2.3",
         "postcss-resolve-nested-selector": "^0.1.1",
         "postcss-selector-parser": "^6.0.13",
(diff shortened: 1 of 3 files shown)


kristerkari:
The only doubt that I have is the `import-lazy` library. Is it safe to use and still maintained (as the last version is from 2019)?

XhmikosR:
The dep is OK, it's just that people move to ESM so there's no need to update it. Stylelint itself is using the same dep for the same reason.

ybiquitous:
FYI. Stylelint has replaced `import-lazy` with dynamic `import()` (https://github.com/stylelint/stylelint/commit/aa9be6f181613c695e80ddf97b0518b1ac674198). This commit will be included in the next major version 16.0.0.

PR #816, opened 2023-05-02
Switch to xo for linting

XhmikosR:
Will conflict with #815, I'll rebase if needed.


src/rules/at-if-no-null/index.js
```diff
@@ -32,20 +32,20 @@ function rule(expectation) {
       }
 
       // If rule != null and (expr), skip
-      if (atrule.params.match(/.* != null and .*/)) {
+      if (/.* != null and .*/.test(atrule.params)) {
```
XhmikosR, line 35: I kept the `test` changes as a separate patch; let me know if you want me to drop them.

src/rules/comment-no-loud/index.js
```diff
@@ -42,7 +42,7 @@ rule.messages = messages;
 rule.meta = meta;
 
 function isLoudComment(comment) {
-  const regex = new RegExp(/^[ \t\n]*\/\*/);
+  const regex = /^[ \t\n]*\/\*/;
```
XhmikosR, line 45: This needs a confirmation just in case

Issue #601, opened 2022-05-07
Add a lint for unnecessary @at-root


nex3:
I've seen a number of cases where authors will write styles like `@at-root .foo &` or even `@keyframes { @at-root from { ... } }`, where the `@at-root` rule is totally unnecessary (in the first case because the selector includes `&` outside of any interpolation, so Sass already knows not to implicitly add a leading `&`; in the second case because `@keyframes` is always bubbled up to the root anyway). It would be useful to have a lint that advises against doing this and suggests deleting the `@at-root` as a fix. Ideally it would work with both the inline selector format *and* the nested format.

PR #846, opened 2023-08-22
Add at-root-no-redundant rule

pamelalozano16:
Fix (#601).

### DIFF
--- a/src/rules/at-root-no-redundant/README.md
+++ b/src/rules/at-root-no-redundant/README.md
@@ -1,0 +1,75 @@
+# at-root-no-redundant
+
+Disallow redundant `@at-root` rule.
+
+<!-- prettier-ignore -->
+```scss
+@at-root .a { margin: 3px; }
+/** ↑
+ * This rule is unnecessary 
+ */
+```
+
+The `@at-root` rule is redundant in the following cases:
+
+- If `@at-root` is already at the root of the document.
+- If any `@at-root` selector contains the parent selector, [`&`](https://sass-lang.com/documentation/style-rules/parent-selector/), outside interpolation.
+- If `@at-root` is nested within a `@keyframes` block.
+
+## Options
+
+### `true`
+
+The following patterns are considered warnings:
+
+<!-- prettier-ignore -->
+```scss
+@at-root .a { margin: 3px; }
+```
+
+<!-- prettier-ignore -->
+```scss
+.a { @at-root .b & { margin: 3px; } }
+```
+
+<!-- prettier-ignore -->
+```scss
+@keyframes slidein {
+  @at-root from {
+    transform: translateX(0%);
+  }
+
+  to {
+    transform: translateX(100%);
+  }
+}
+```
+
+The following patterns are _not_ considered warnings:
+
+<!-- prettier-ignore -->
+```scss
+.a { @at-root .b { margin: 3px; } }
+```
+
+<!-- prettier-ignore -->
+```scss
+.a { @at-root .b#{&} { margin: 3px; } }
+```
+
+<!-- prettier-ignore -->
+```scss
+.a {
+  @at-root .b {
+    @keyframes slidein {
+      from {
+        transform: translateX(0%);
+      }
+    
+      to {
+        transform: translateX(100%);
+      }
+    }
+  }
+}
+```
--- a/src/rules/at-root-no-redundant/__tests__/index.js
+++ b/src/rules/at-root-no-redundant/__tests__/index.js
@@ -1,0 +1,136 @@
+"use strict";
+
+const { messages, ruleName } = require("..");
+
+testRule({
+  ruleName,
+  config: [true],
+  customSyntax: "postcss-scss",
+
+  accept: [
+    {
+      code: `
+      .a { @at-root .b { c: d } }
+      `,
+      description: "@at-root rule is nested."
+    },
+    {
+      code: `
+      .a { @at-root .b#{&} { c: d; } }
+      `,
+      description:
+        "@at-root is followed by a selector with an interpolated `&`."
+    },
+    {
+      code: `
+      .a { @at-root .b { .c & { d: e; } } }
+      `,
+      line: 2,
+      column: 12,
+      message: messages.rejected,
+      description: "A parent selector (&) is used under the @at-root rule."
+    },
+    {
+      code: `
+      .a {
+        @at-root .b {
+          @keyframes slidein {
+            from {
+              transform: translateX(0%);
+            }
+          
+            to {
+              transform: translateX(100%);
+            }
+          }
+        }
+      }`,
+      description: "@at-root outside a keyframes block, nested."
+    },
+    {
+      code: `
+      @mixin foo {}
+      
+      @keyframes slidein {
+        from {
+          @include foo;
+          transform: translateX(0%);
+        }
+      
+        to {
+          transform: translateX(100%);
+        }
+      }`,
+      description: "Other at-rules nested inside @keyframes block."
+    }
+  ],
+  reject: [
+    {
+      code: `
+      @at-root .a { b: c; }
+      `,
+      line: 2,
+      column: 7,
+      message: messages.rejected,
+      description: "@at-root rule is already in the root."
+    },
+    {
+      code: `
+      .a { @at-root .b & { c: d; } }
+      `,
+      line: 2,
+      column: 12,
+      message: messages.rejected,
+      description:
+        "@at-root is followed by a selector with a `&` outside an interpolation."
+    },
+    {
+      code: `
+      .a { @at-root .b & #{.c} { d: e; } }
+      `,
+      line: 2,
+      column: 12,
+      message: messages.rejected,
+      description:
+        "@at-root is followed by a selector with a `&` outside an interpolation."
+    },
+    {
+      code: `
+      @keyframes slidein {
+        @at-root from {
+          transform: translateX(0%);
+        }
+      
+        to {
+          transform: translateX(100%);
+        }
+      }`,
+      line: 3,
+      column: 9,
+      message: messages.rejected,
+      description: "@at-root inside a keyframes block, top level."
+    },
+    {
+      code: `
+      @keyframes slidein {
+        from {
+          transform: translateX(0%);
+             @at-root from {
+              transform: translateX(100%);
+             }
+             to {
+                transform: translateX(0%);
+              }
+        }
+      
+        to {
+          transform: translateX(100%);
+        }
+      }`,
+      line: 5,
+      column: 14,
+      message: messages.rejected,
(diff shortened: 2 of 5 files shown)


src/rules/at-root-no-redundant-selector/index.js
```diff
@@ -0,0 +1,74 @@
...
+const ruleUrl = require("../../utils/ruleUrl");
+
+const ruleName = namespace("at-root-no-redundant-selector");
+
+function isInterpolated(string) {
+  const interpolationRegex = /#{(.*?)}/g;
+  return interpolationRegex.test(string);
+}
+
+function isAtRootRuleNested(node) {
```
Goodwine, line 14: You don't have to manually go down the tree recursively, you can use `walkAtRules` with any container node, which inclides AtRule nodes, so.. you can do this instead:

```js
function findAtRoot(node) {
  let atRoot;
  node.walkAtRules('at-root', node => {
    atRoot = node;
    return false;
  });
  return atRoot;
}
```

src/rules/at-root-no-redundant-selector/index.js
```diff
@@ -0,0 +1,74 @@
...
+          message: messages.rejected,
+          node: decl,
+          result,
+          ruleName
+        });
+      }
+      if (decl.name === "keyframes" && isAtRootRuleNested(decl)) {
+        utils.report({
+          message: messages.rejected,
+          node: decl,
```
Goodwine, line 61: You don't want to report the error on the `@keyframes`, instead you want to report it on the `@at-rule` node.

You can implement the suggestion from above with `findAtRoot`, and report it there, e.g.:
   ```js
   if (node.name === "keyframes") {
     const atRoot = findAtRoot(node);
     if (atRoot) {
       utils.report({
         message: messages.rejected,
         node: atRoot,
         result,
         ruleName
       });
     }
   }
   ```

src/rules/at-root-no-redundant-selector/index.js
```diff
@@ -0,0 +1,74 @@
...
+  return (root, result) => {
+    const validOptions = utils.validateOptions(result, ruleName, {
+      actual
+    });
+
+    if (!validOptions) {
+      return;
+    }
+
+    root.walkAtRules(decl => {
```
Goodwine, line 44: Optional: you can pass a filter to `walkAtRules`, for example:

```js
root.walkAtRules('at-root', node => { ... })
```

Doing so, you can omit the checks for `node.name === 'at-root'` because it would be guaranteed to be this node. However, you won't be able to match `@keyframes`. But.. you can always iterate up, for example, if I want to check whether an at-root exists within a `@keyframes` node, I could do:

```js
function isWithinKeyframes(node) {
  let parent = node.parent;
  while (parent != null) {
    if (parent.type === 'at-rule' && parent.name === 'keyframes') {
      return true;
    }
    parent = node.parent;
  }
  return false;
}
```

With this, your overall implementation could be greatly reduced to:

```js
function rule(actual) {
  return (root, result) => {
    const validOptions = utils.validateOptions(result, ruleName, {
      actual
    });

    if (!validOptions) {
      return;
    }

    root.walkAtRules('at-root', node => {
      if (
        node.parent.type === 'root' ||
        node.params.replace(/#{.*}/g, '').includes('&') ||
        isWithinKeyframes(node)) {
          utils.report({
            message: messages.rejected,
            node,
            result,
            ruleName
          });
        }
    });
}
```

Goodwine (changes requested):

src/rules/at-root-no-redundant-selector/README.md
```diff
@@ -0,0 +1,67 @@
+# at-root-no-redundant-selector
```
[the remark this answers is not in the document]
pamelalozano16: I had initially suggested `no-redundant-at-root` as well, @nex3 suggested `at-root-no-redundant-selector`, thoughts? 🤔

src/rules/at-root-no-redundant-selector/__tests__/index.js
```diff
@@ -0,0 +1,129 @@
...
+
+testRule({
+  ruleName,
+  config: [true],
+  customSyntax: "postcss-scss",
+
+  accept: [
+    {
+      code: `
+      .a { @at-root .b: c; }
```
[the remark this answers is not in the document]
pamelalozano16: Oh right, I'll verify all of them thanks!! :)

src/rules/at-root-no-redundant-selector/__tests__/index.js
```diff
@@ -0,0 +1,129 @@
...
+      `,
+      description: "@at-root rule is nested."
+    },
+    {
+      code: `
+      .a { @at-root b#{&}: { c: d; } }
```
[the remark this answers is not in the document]
pamelalozano16: Oh right, I was searching for an interpolation with any character instead of an interpolation with &.

src/rules/at-root-no-redundant-selector/index.js
pamelalozano16: Great suggestion, thank you!!

src/rules/at-root-no-redundant-selector/README.md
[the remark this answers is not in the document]
Goodwine: Maybe we can discuss this today at coworking time, I still think that `no-redundant-at-root` is a more accurate name, since the selector is not the problem, even less so when the `@at-root` is already at the root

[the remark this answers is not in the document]
Goodwine: discussed this briefly, the id should indeed start with `at-root` following the same pattern as other checks, so I think then the ID should be `at-root-no-redundant`, without the "selector" part 

ultimately the final say will be from the maintainer :)

src/rules/at-root-no-redundant-selector/index.js
```diff
@@ -0,0 +1,74 @@
...
+      if (decl.type !== "atrule") return;
+      if (
+        decl.name === "at-root" &&
+        (decl.parent.type === "root" ||
+          (decl.params.includes("&") && !isInterpolated(decl.params)))
```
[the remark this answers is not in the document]
pamelalozano16: Solved this by matching the regex to `/#{[&]}/g`

src/rules/at-root-no-redundant-selector/index.js
pamelalozano16: Awesome!! Thanks

src/rules/at-root-no-redundant/README.md
```diff
@@ -0,0 +1,67 @@
...
+The following patterns are considered warnings:
+
+<!-- prettier-ignore -->
+```scss
+@at-root .a { margin: 3px; }
+```
+
+<!-- prettier-ignore -->
+```scss
+.a { @at-root .b & : { margin: 3px; } }
```
Goodwine, line 24:
```suggestion
.a { @at-root .b & { margin: 3px; } }
```

src/rules/at-root-no-redundant/README.md
```diff
@@ -0,0 +1,67 @@
...
+The following patterns are _not_ considered warnings:
+
+<!-- prettier-ignore -->
+```scss
+.a { @at-root .b { margin: 3px; } }
+```
+
+<!-- prettier-ignore -->
+```scss
+.a { @at-root .b#{&}: { margin: 3px; } }
```
Goodwine, line 49:
```suggestion
.a { @at-root .b#{&} { margin: 3px; } }
```

src/rules/at-root-no-redundant/__tests__/index.js
```diff
@@ -0,0 +1,141 @@
...
+      }`,
+      description: "@at-root outside a keyframes block, nested."
+    },
+    {
+      code: `
+      @mixin reset-list {
+        margin: 0;
+        padding: 0;
+        list-style: none;
+      }
```
Goodwine, line 56: Nit: this doesn't have to have any contents to assert the behavior being tested, it can be simplified to `@mixin foo {}`

src/rules/at-root-no-redundant/__tests__/index.js
```diff
@@ -0,0 +1,141 @@
...
+             to {
+                transform: translateX(0%);
+              }
+        }
+      
+        to {
+          transform: translateX(100%);
+        }
+      }
+          `,
```
Goodwine, line 134:
```suggestion
      }`,
```

src/rules/at-root-no-redundant/index.js
```diff
@@ -0,0 +1,59 @@
...
+
+const { utils } = require("stylelint");
+const namespace = require("../../utils/namespace");
+const ruleUrl = require("../../utils/ruleUrl");
+
+const ruleName = namespace("at-root-no-redundant");
+
+function isWithinKeyframes(node) {
+  let parent = node.parent;
+  while (parent.type !== "root") {
```
Goodwine, line 11: Nit: this condition is a little bit risky as it could throw an NPE if the node is "orphan". In practice it probably doesn't matter, but a safer check would be `while(parent)` or to be more explicit `while(parent != null)`

Goodwine (approved):
LGTM besides the remaining comments

pamelalozano16:
@kristerkari please let us know your thoughts :)

kristerkari:
Thanks! I'll need to go through all the new PRs 👍

src/rules/at-root-no-redundant/README.md
```diff
@@ -0,0 +1,67 @@
+# at-root-no-redundant
+
+Ban unnecessary `@at-root` rule.
```
kristerkari, line 3: There are some small things to point out related to how the README and message should be formatted. 

Please check Stylelint's developer guide for more info:
https://stylelint.io/developer-guide/rules#write-the-readme

```md
The single-line description is in the form of:

    - "Disallow ..." for no rules
    - "Limit ..." for max rules
    - "Require ..." for rules that accept "always" and "never" options
    - "Specify ..." for everything else
```

src/rules/at-root-no-redundant/index.js
```diff
@@ -0,0 +1,59 @@
...
+    if (parent.type === "atrule" && parent.name === "keyframes") {
+      return true;
+    }
+    parent = parent.parent;
+  }
+  return false;
+}
+
+const messages = utils.ruleMessages(ruleName, {
+  rejected: "Unnecessary @at-root rule"
```
kristerkari, line 21: If I'm not mistaken this should be `Unexpected @at-root rule`.

https://stylelint.io/developer-guide/rules#add-problem-messages

pamelalozano16: Right, thank you!

kristerkari (approved):

PR #847, opened 2023-08-24
Add property-no-unknown rule

### LINKED ISSUE #490: Add a Sass-specific property-no-unknown
It would be useful to have a version of stylelint's [property-no-unknown](https://stylelint.io/user-guide/rules/property-no-unknown) rule that's aware of Sass's [nested properties](https://sass-lang.com/documentation/style-rules/declarations#nesting), and so can accurately recognize that

```scss
a {
  font: {weight: bold}
}
```

is valid but

```scss
a {
  font: {stuff: bold}
}
```

is invalid.

pamelalozano16:
Fix #490.


src/utils/isStandardSyntaxDeclaration.js
```diff
@@ -0,0 +1,49 @@
+"use strict";
+
+/**
+ * Check whether a declaration is standard
+ *
+ * @param {import('postcss').Declaration | import('postcss-less').Declaration} decl
```
jathak, line 6: I believe adding this would mean that postcss-less has to be a non-dev dependency, which we wouldn't want

kristerkari:
I guess that the only difference from the standard stylelint rule is the support for nested property names?

The README could also mention that this rule should be used instead of the stylelint one, e.g. https://github.com/stylelint-scss/stylelint-scss/blob/master/src/rules/at-rule-no-unknown/README.md#at-rule-no-unknown

kristerkari (approved):

PR #848, opened 2023-08-28
Add at-use-no-redundant-alias rule

### LINKED ISSUE #445: Add a lint for an explicit namespace that's the same as the default
If a user writes something like `@use "foo" as foo`, the `as foo` is totally redundant since it's the same as the default namespace that the `@use` rule would provide on its own. It would be nice to have a lint for this case.

pamelalozano16:
Fix #445.

### DIFF
--- a/src/rules/at-use-no-redundant-alias/README.md
+++ b/src/rules/at-use-no-redundant-alias/README.md
@@ -1,0 +1,44 @@
+# at-use-no-redundant-alias
+
+Disallow redundant namespace aliases.
+
+```scss
+@use "foo" as foo;
+/**           ↑
+ * Disallow this */
+```
+
+By default, the [module's namespace](https://sass-lang.com/documentation/at-rules/use/#loading-members)
+is the last component of the module’s URL.
+
+## Options
+
+### `true`
+
+The following patterns are considered warnings:
+
+```scss
+@use "foo" as foo;
+```
+
+```scss
+@use "sass:math" as math;
+```
+
+```scss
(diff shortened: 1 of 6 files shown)


src/rules/at-use-no-redundant-alias/README.md
```diff
@@ -0,0 +1,38 @@
+# at-use-no-redundant-alias
+
+By default, the module's namespace is just the last component of the module’s URL.
```
kristerkari, line 3: Please check the format for the rule description from the developer guide:
https://stylelint.io/developer-guide/rules#write-the-readme

src/rules/at-use-no-redundant-alias/index.js
```diff
@@ -0,0 +1,53 @@
+"use strict";
+
+const { utils } = require("stylelint");
+const namespace = require("../../utils/namespace");
+const ruleUrl = require("../../utils/ruleUrl");
+
+const ruleName = namespace("at-use-no-redundant-alias");
+
+const messages = utils.ruleMessages(ruleName, {
+  rejected: "Unnecessary explicit namespace"
```
kristerkari, line 10: Check the developer guide for the format:
https://stylelint.io/developer-guide/rules#add-problem-messages

src/rules/at-use-no-redundant-alias/README.md
pamelalozano16: Thanks !!

kristerkari (approved):

PR #849, opened 2023-08-29
Add function-calculation-no-interpolation rule

### LINKED ISSUE #539: Add a rule that forbids interpolation in calculations
The upcoming [first-class `calc()` feature](https://sass-lang.com/blog/request-for-comments-first-class-calc) adds the ability to refer to variables and call functions directly within `calc()`, `clamp()`, `min()`, and `max()` expressions (collectively referred to as "calculations"). Although it's still possible to inject values into these functions using interpolation, once first-class `calc()` is available there won't be any need to do so and in fact it will make the stylesheet more error-prone. It would be nice to have a lint that guides users away from this practice and towards using variables and functions directly.

pamelalozano16:
Fix #539.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 5.2.0
+
+- Added: `function-calculation-no-interpolation` rule to forbid interpolation in calc functions (#539).
+
+**Full Changelog**: https://github.com/stylelint-scss/stylelint-scss/compare/v5.1.0...v5.2.0
+
 # 5.1.0
 
 - Added: `function-disallowed-list` rule support to ban specific built-in functions (#422, #844).
--- a/README.md
+++ b/README.md
@@ -30,9 +30,7 @@ Create the `.stylelintrc.json` config file (or open the existing one), add `styl
 
 ```jsonc
 {
-  "plugins": [
-    "stylelint-scss"
-  ],
+  "plugins": ["stylelint-scss"],
   "rules": {
     // recommended rules
     "at-rule-no-unknown": null,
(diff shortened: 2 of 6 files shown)


src/rules/function-no-interpolation/README.md
```diff
@@ -0,0 +1,61 @@
+# function-no-interpolation
+
+The upcoming [first-class `calc()` feature](https://sass-lang.com/blog/request-for-comments-first-class-calc/) adds the ability to refer to variables and call functions directly within `calc()`, `clamp()`, `min()`, and `max()` expressions.
```
jathak, line 3: This feature has now launched. It would also be good to connect this to the rationale for this lint as described in the issue

src/rules/function-no-interpolation/__tests__/index.js
```diff
@@ -0,0 +1,82 @@
...
+    },
+    {
+      code: `
+      $c: 1;
+      .a { .b: min(#{$c} + 1px); }
+      `,
+      line: 3,
+      column: 12,
+      message: messages.rejected("min"),
+      description: "`max` function with interpolation"
```
kristerkari, line 69: This should be `min` instead of `max`.

src/rules/function-no-interpolation/README.md
```diff
@@ -0,0 +1,64 @@
+# function-no-interpolation
+
+Since the release of [first-class `calc()`](https://sass-lang.com/documentation/values/calculations/),
```
kristerkari, line 3: Please check the developer guide for REAME format. You can also have a look at the existing stylelint rules:

- https://stylelint.io/developer-guide/rules#write-the-readme
- https://github.com/stylelint/stylelint/tree/main/lib/rules

pamelalozano16: Awesome, thanks for the resources, just fixed the README.md files in all PRs. 
Although.. I didn't find a rule similar to this rule (maybe [`function-calc-no-unspaced-operator`](https://github.com/stylelint/stylelint/tree/main/lib/rules/function-calc-no-unspaced-operator)?). Please let me know if I'm missing something. 
Thanks! :)

pamelalozano16: @kristerkari Any thoughts on this ? :)

kristerkari: Thanks for the fixes @pamelalozano16 ! 

One thing that I was still thinking about this `function-no-interpolation` rule: the name of this rule suggests that it could apply to all functions, but it only applies to the "calculation" functions, so maybe that should be part of the name to avoid possible confusion?

kristerkari: The other PRs looked good now, so I can merge + cut a new release once this PR gets fixed.

kristerkari: Maybe `function-calculation-no-interpolation` could a good name?

pamelalozano16: Done! :)

src/rules/calc-functions-no-interpolation/README.md
```diff
@@ -0,0 +1,64 @@
+# calc-functions-no-interpolation
```
kristerkari, line 1: @pamelalozano16 Sorry for not being clear about it, but this naming style won't work. 

If you look at the stylelint naming, it's done like this: `function-type-no-something`, e.g. `function-calc-no-unspaced-operator`.

Screenshot 2023-09-13 at 23 34 02

kristerkari: @pamelalozano16 the CSS spec seems to refer to the calc and other functions as "math-functions" (https://drafts.csswg.org/css-values/#math-function), but using that in the name might be confusing as Sass has it's own math functions

kristerkari: Both the Sass docs and the CSS spec are referring to calculation functions, so maybe that's the name that would make most sense.

kristerkari: So maybe the name could be something like `function-calculation-no-interpolation`.

Using `function-calc-no-interpolation` would be another option, but that would suggest that it would only work with the `calc()` function.

pamelalozano16: Thanks for the clarification! `calculation` makes sense to me since both Sass docs and the CSS spec refer to these as calculation functions as you mentioned. :)

kristerkari (approved):

kristerkari:
@pamelalozano16  Thanks for the contributions! The rules are no released in the 5.2.0 version.

PR #857, opened 2023-09-15
fix: add missing `known-css-properties` dependency

G-Rath:
#847 introduced a dependency on `known-css-properties` but didn't declare that in `package.json`


kristerkari:
Thanks! Good catch.

PR #858, opened 2023-09-21
Add `stylelintType: "parseError"` when `parseSelector` fails

wlewis-formative:
Without this, I believe parse errors are being treated as warnings by `stylelint`. Updating this to match e.g. the [`parseSelector`](https://github.com/stylelint/stylelint/blob/74425d19a58449471498a83da519346290defee5/lib/utils/parseSelector.js#L16) function in `stylelint` causes them to be treated as errors.

## To reproduce

- `git clone git@github.com:wlewis-formative/stylelint-scss-min-repro.git`
- Run `npm lint`
- Run `echo $?`

The exit code is `0`, even though errors were reported:

```shell
$ npm run lint

> stylelint-scss-min-repro@0.0.1 lint
> stylelint Bad.tsx


Bad.tsx
 4:3  ✖  Cannot parse selector

1 problem (1 error, 0 warnings)

$ echo $?
0
```

This issue occurs because `postcss-styled-syntax` is parsing inline comments as selectors instead of reporting an error. The error is only discovered later (here).

### DIFF
--- a/src/utils/parseSelector.js
+++ b/src/utils/parseSelector.js
@@ -6,6 +6,6 @@ module.exports = function (selector, result, node, cb) {
   try {
     return postCssSelectorParser(cb).processSync(selector);
   } catch (e) {
-    result.warn("Cannot parse selector", { node });
+    result.warn("Cannot parse selector", { node, stylelintType: "parseError" });
   }
 };


kristerkari:
@wlewis-formative Thanks for this one! I have just one question in mind: do you know if this change is backwards compatible to the current minimum supported version (`14.5.1`) of this package?

wlewis-formative:
I _think_ so. I checked out the commit with the tag `14.5.1` (`git checkout tags/14.5.1`) in the `stylelint` repo, and both `parseSelector` and `transformSelector` call `result.warn` with `stylelintType: 'parseError'`:

```javascript
} catch {
  result.warn('Cannot parse selector', { node, stylelintType: 'parseError' });
  //                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
}
```

kristerkari:
@wlewis-formative Great, thanks for checking! I think that this is good to go then 👍

PR #860, opened 2023-09-27
Add `--fix` to `at-root-no-redundant`.

pamelalozano16:
Add fix to `at-root-no-redundant` by removing the redundant `@at-root` rule from the node string and parsing / appending the new node.


kristerkari:
Thanks!

PR #866, opened 2023-10-02
Add `--fix` to `at-use-no-redundant-alias`

pamelalozano16:
Add fix to `at-use-no-redundant-alias` by removing the redundant alias from the node string and parsing / appending the new node.


pamelalozano16:
@kristerkari Any thoughts? :) Also, could we make a new release after this PR is merged?

kristerkari:
Yes, sorry for the delay. I've been crazy busy.

Looks good 👍

PR #867, opened 2023-10-04
Add `load-no-partial-leading-underscore` rule.

### LINKED ISSUE #859: Deprecate `at-import-no-partial-leading-underscore` and add `load-no-partial-leading-underscore`.
The same rules in [at-import-no-partial-leading-underscore](https://g3doc.corp.google.com/third_party/javascript/node_modules/stylelint_scss/v5_2_1/src/rules/at-import-no-partial-leading-underscore/README.md?cl=head) of disallowing leading underscores and file extensions apply to [@use](https://sass-lang.com/documentation/at-rules/use/), [@forward](https://sass-lang.com/documentation/at-rules/forward/) and [meta.load-css()](https://sass-lang.com/documentation/modules/meta/#mixins) url parameter. The existing `at-import-no-partial-leading-underscore` rule could be expanded to cover all these cases and then be renamed to `load-no-partial-leading-underscore`.

pamelalozano16:
Fix #859.

### DIFF
--- a/src/rules/at-import-no-partial-leading-underscore/index.js
+++ b/src/rules/at-import-no-partial-leading-underscore/index.js
@@ -11,7 +11,8 @@ const messages = utils.ruleMessages(ruleName, {
 });
 
 const meta = {
-  url: ruleUrl(ruleName)
+  url: ruleUrl(ruleName),
+  deprecated: true
 };
 
 function rule(actual) {
@@ -21,6 +22,16 @@ function rule(actual) {
     if (!validOptions) {
       return;
     }
+
+    result.warn(
+      "'at-import-no-partial-leading-underscore' has been deprecated, " +
+        "and will be removed in '6.0'. Use 'load-no-partial-leading-underscore' instead.",
+      {
+        stylelintType: "deprecation",
+        stylelintReference:
+          "https://github.com/stylelint-scss/stylelint-scss/blob/v5.2.1/src/rules/at-import-no-partial-leading-underscore/README.md"
+      }
+    );
 
     function checkPathForUnderscore(path, decl) {
       // Stripping trailing quotes and whitespaces, if any
--- a/README.md
+++ b/README.md
@@ -80,7 +80,6 @@ Please also see the [example configs](./docs/examples/) for special cases.
 
 ### `@`-import
 
-- [`at-import-no-partial-leading-underscore`](./src/rules/at-import-no-partial-leading-underscore/README.md): Disallow leading underscore in partial names in `@import`.
 - [`at-import-partial-extension`](./src/rules/at-import-partial-extension/README.md): Require or disallow extension in `@import` commands (Autofixable).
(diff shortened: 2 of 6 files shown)


src/rules/at-import-no-partial-leading-underscore/index.js
```diff
@@ -16,6 +16,16 @@ const meta = {
 
 function rule(actual) {
   return (root, result) => {
+    result.warn(
+      "'at-import-no-partial-leading-underscore' has been deprecated, " +
+        "and will be removed in '6.0'. Use 'load-no-partial-leading-underscore' instead.",
+      {
+        stylelintType: "deprecation",
+        stylelintReference:
+          "https://github.com/stylelint-scss/stylelint-scss/tree/master/src/rules/load-no-partial-leading-underscore/README.md"
```
kristerkari, line 25: This deprecation message looks nice, but just a couple of small things to fix.

From the developer guide:

> Deprecating rules doesn't happen very often. When you do, you must:
>
> 1. Point the `stylelintReference` link to the specific version of the rule README on the GitHub website, so that it is always accessible.
> 2. Add the appropriate metadata to mark the rule as deprecated like `rule.meta = { deprecated: true }`.

`stylelintReference`  should probably point to the old rule, not the new one, and the url should be locked to a version (e.g. current release). You can use this for reference:
https://github.com/stylelint/stylelint/blob/13.7.0/lib/rules/at-rule-blacklist/index.js#L32C1-L35C6

Also,`deprecated: true` needs to be added to the `meta` object in the rule.

src/rules/at-import-no-partial-leading-underscore/index.js
```diff
@@ -16,6 +16,16 @@ const meta = {
 
 function rule(actual) {
   return (root, result) => {
+    result.warn(
```
kristerkari, line 19: Looking at this, it seems that the call to `result.warn` should be placed to be after the `validOptions` check:
https://github.com/stylelint/stylelint/blob/13.7.0/lib/rules/at-rule-blacklist/index.js#L32C1-L35C6

kristerkari:
Could you please also edit the main README of this repo? 

1.   The `at-import-no-partial-leading-underscore` rule needs to be marked as `(deprecated)` in the README listing
2. A new "Load" category needs to be added and the `load-no-partial-leading-underscore` rule under it.

src/rules/at-import-no-partial-leading-underscore/index.js
pamelalozano16: Right, thanks!

src/rules/at-import-no-partial-leading-underscore/index.js
pamelalozano16: Gotcha, sorry I missed that! Thank you

pamelalozano16:
@kristerkari Thanks for the review! All done. I added a section of [Deprecated rules](https://github.com/stylelint-scss/stylelint-scss/blob/080b633d017a3c3f6467ec5bf63a295f25d752f9/README.md#deprecated), let me know if that looks ok please. Thanks!

kristerkari (approved):
Thanks! Looks good now 👍  

I'll do the release tomorrow as I need to go to sleep now 🙂

PR #870, opened 2023-10-17
Check if all selectors followed by @at-root include '&' outside interpolation

### LINKED ISSUE #868: at-rule-no-redundant triggers incorrectly when some (but not all) selector has "&"
Example input in Sass:
```scss
.foo {
  @at-root .bar,
  &baz {
    a: b;
  }
}

.foo {
  .bar,
  &baz {
    a: b;
  }
}
```
https://sass-lang.com/playground/#eJwzNNRLy89XqOZSUHBILNEtys8vUdBLSizSAQqoJSVWgWUUFBKtFJKsgaxarlouLrgO/AoBMc4WWQ==

The rule for `@at-root` will trigger saying it's redundant and should be removed. This implies both rules have the same output, however that's not correct. So rather than showing the linter error when _any_ selector has `&`, it should happen only when `_every_` selector does.

cc: @pamelalozano16

pamelalozano16:
Fix #868.

### DIFF
--- a/src/rules/at-root-no-redundant/__tests__/index.js
+++ b/src/rules/at-root-no-redundant/__tests__/index.js
@@ -62,6 +62,18 @@ testRule({
         }
       }`,
       description: "Other at-rules nested inside @keyframes block."
+    },
+    {
+      code: `
+      .foo {
+        @at-root .bar,
+        &baz {
+          a: b;
+        }
+      }
+      `,
+      description:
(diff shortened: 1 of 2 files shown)


kristerkari:
Looks good, is it possible to also add the autofix for this case?

pamelalozano16:
@kristerkari This rule already contains an autofix. :) I just added a test to make sure it works in this case as well, please take a look and let me know if it's ok.

kristerkari:
Looks good 👍

PR #872, opened 2023-10-19
Add `block-no-redundant-nesting` rule

### LINKED ISSUE #679: Rule suggestion: `block-no-unnecessary-child-block`
I suggest a new Stylelint rule that reports a single block inside another block:

```scss
// BAD
.foo {
    .bar { … }
}

// GOOD
.foo .bar { … }
```

```scss
// BAD
.foo {
    &-bar { … }
}

// GOOD
.foo-bar { … }
```

```scss
// GOOD, there are multiple child blocks
.foo {
    .bar { … }
    &-baz { … }
}
```

```scss
// GOOD, the outer selector contains a comma
.foo,
.bar {
    &-baz { … }
}
```

FloEdelmann:
Fixes #679.


src/rules/block-no-unnecessary-nesting/__tests__/index.js
```diff
@@ -0,0 +1,183 @@
...
+              color: red;
+            }
+          }
+        }
+      `,
+      fixed: `
+        .foo .bar {
+        }
+        .foo-bar-baz {
+              color: red
```
kristerkari, line 154: Small thing, but it would nice if the autofix would indent the properties correctly :)

src/rules/block-no-unnecessary-nesting/index.js
```diff
@@ -0,0 +1,77 @@
+"use strict";
+
+const resolveNestedSelector = require("postcss-resolve-nested-selector");
+const { utils } = require("stylelint");
+const namespace = require("../../utils/namespace");
+const ruleUrl = require("../../utils/ruleUrl");
+
+const ruleName = namespace("block-no-unnecessary-nesting");
```
kristerkari, line 8: Hmm.. it looks like stylelint prefers to use `no-redundant` for the rules instead of `no-unnecessary`, so it would be better to follow the same convention.

kristerkari:
Thanks! Looks good, just a couple of small things that I commented on :)

src/rules/block-no-unnecessary-nesting/index.js
FloEdelmann: Renamed in [`d01ccc0`](https://github.com/stylelint-scss/stylelint-scss/pull/872/commits/d01ccc0afbae9d539ca634715bd5655ee40d33d1).

src/rules/block-no-unnecessary-nesting/__tests__/index.js
kristerkari: Well, actually, the core stylelint package has removed all styling related rules, so maybe the indentation is something that we want to leave to tools like prettier instead (check https://github.com/stylelint/stylelint/blob/main/docs/migration-guide/to-15.md#deprecated-stylistic-rules)

kristerkari:
One probablem that I noticed when reviewing this rule is that the rule does not seem to take into account that Sass supports nesting CSS properties.

This `accepted` test should pass:

```js
    {
      code: `
        .foo {
          font: {
            size: 16px;
          }
        }
      `,
      description: "Sass' nested properties"
    }
```

...as it compiles to:

```css
.foo {
  font-size: 16px;
}
```

kristerkari (approved):
Thanks! Looks really good now 👍 

I'll release this in the next release (currently missing only #867)

PR #898, opened 2023-12-11
Support Stylelint 16

### LINKED ISSUE #894: Support stylelint16
Support stylelint16 https://stylelint.io/migration-guide/to-16/

kristerkari:
Fixes #894

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,7 +10,7 @@ on:
 
 env:
   FORCE_COLOR: 2
-  NODE: 18
+  NODE: 21
 
 permissions:
   contents: read
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,7 +10,7 @@ on:
 
 env:
   FORCE_COLOR: 2
-  NODE: 21
+  NODE: 20
 
 permissions:
   contents: read
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: [14, 16, 18, 20]
+        node: [21]
         os: [ubuntu-latest, windows-latest]
 
     steps:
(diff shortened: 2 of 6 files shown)


.github/workflows/lint.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 env:
   FORCE_COLOR: 2
-  NODE: 18
+  NODE: 21
```
ybiquitous, line 13: JFYI. For linting, using `latest` or `node` may be better, instead of a specific Node.js version.
See https://github.com/actions/setup-node#supported-version-syntax

kristerkari: Hmm..  

Wouldn't that mean that the CI could break when a new Node.js version with breaking changes gets released and the project is not yet compatible with it?

kristerkari:
> Since Node.js 21 will end before Node.js 20, using `20` rather than `21` may be better.

The problem that I had with Node.js 20 was that the test suite was crashing with it. I would imagine that it will do the same in CI.

kristerkari:
Actually, it does not seem to crash with Node v20.10.0.

I guess that my Node.js 20 was just a bit outdated.

ybiquitous:
I also had trouble in Jest when testing the ESM codebase, e.g. segmentation fault. 😓 

If you're not particular about Jest, [`stylelint-test-rule-node`](https://www.npmjs.com/package/stylelint-test-rule-node) is one of the alternatives as the [migration guide](https://stylelint.io/migration-guide/to-16#plugins) said.

kristerkari:
> If you're not particular about Jest, [`stylelint-test-rule-node`](https://www.npmjs.com/package/stylelint-test-rule-node) is one of the alternatives as the [migration guide](https://stylelint.io/migration-guide/to-16#plugins) said.

For now it should be enough to just add support for Stylelint 16. 

I know that the test suite and ESM migrations are something that are on the roadmap for a future version.

ybiquitous:
Although Node.js Test Runner still supports coverage experimentally.
See https://nodejs.org/api/test.html#collecting-code-coverage

ybiquitous (approved):
LGTM 👍🏼

PR #902, opened 2023-12-12
Fix `at-import-partial-extension` end positions

ybiquitous:
Part of #652

### DIFF
--- a/src/rules/at-import-partial-extension/index.js
+++ b/src/rules/at-import-partial-extension/index.js
@@ -2,6 +2,7 @@
 
 const nodeJsPath = require("path");
 const { utils } = require("stylelint");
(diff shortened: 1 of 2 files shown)


src/rules/at-import-partial-extension/index.js
```diff
@@ -82,15 +84,20 @@ function rule(expectation, _, context) {
...
+          const index =
+            atRuleParamIndex(atRule) + atRule.params.lastIndexOf(dotExt);
+
           utils.report({
             message: messages.rejected(extension),
-            node: decl,
-            word: extension,
+            node: atRule,
+            index,
+            endIndex: index + dotExt.length,
```
ybiquitous, line 100: [note] I think the end position should include a dot (`.`) since the message also includes it:

```
Unexpected extension ".scss" in @import
```

PR #947, opened 2024-01-11
Allow calc function interpolation when declaring CSS custom properties.

pamelalozano16:
In the example case `--test: calc(#{$c} + 1);` , `#{$c}`  is mistakenly identified as an error. Interpolation is used in SassScript to inject dynamic values into a [CSS custom property](https://sass-lang.com/documentation/style-rules/declarations/#custom-properties). This allows interpolation when declaring CSS variables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 6.0.1
+
+- `function-calculation-no-interpolation` allows calc function interpolation in SassScript when declaring CSS custom properties.
+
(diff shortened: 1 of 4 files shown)


kristerkari:
Thanks! I wonder if we should ignore the custom properties by default, or put this behind an option?

pamelalozano16:
@kristerkari it should _always_ ignore custom properties since these passed through to CSS as-is, for any SassScript interpolation is required.

kristerkari:
got it, thanks :)

PR #948, opened 2024-01-16
property-no-unknown for deeper nested properties

pamelalozano16:
Fix: `scss/property-no-unknown` is not recognizing more than two-level nested properties.
E.g.
```
background: {
  position: {
    x: left;
  }
}
```
compiles to: `background-position-x: left;`. This is a [valid property](https://developer.mozilla.org/en-US/docs/Web/CSS/background-position-x).

However, it's currently raising the following error: `Unexpected unknown property "position-x" (scss/property-no-unknown)`.


kristerkari:
Thanks!

PR #952, opened 2024-01-22
Automatic fix for unused private members

pamelalozano16:


kristerkari:
Thanks!

I think that it's a bit tricky to add autofix for this rule. The autofix will delete styles, which might not always be what the user expects to happen.

Should something like this be instead implemented with an extra option that you can opt-in to using when you want the styles to be automatically fixed?

pamelalozano16:
Gotcha, what about a config option like `allowAutoFix: true` ?

kristerkari:
> what about a config option like `allowAutoFix: true`

I was trying to search stylelint's repo to see if they have anything similar there, but it seems that there are no autofixes like this (that remove code). I'm not even 100% sure if adding an autofix like this is a good idea since it needs to be enabled with a secondary option.

Another option for the option name could be `enableAutoFix`.

pamelalozano16:
@kristerkari Just added enableAutoFix as a secondary option

PR #958, opened 2024-02-21
Rename rules to use inclusive language

### LINKED ISSUE #571: Rename rules for blacklist and whitelist
Stylelint 13.7 has renamed rules using `-whitelist` and `-blacklist` in their name (keeping the old rule names as deprecated until the next major version of course): https://stylelint.io/CHANGELOG#1370

It might make sense to do a similar change in stylelint-scss

01taylop:
As well as using inclusive language, this change aligns with similar improvements made in Stylelint: https://stylelint.io/CHANGELOG#1370.

- `at-import-partial-extension-whitelist` -> `at-import-partial-extension-allowed-list`
- `at-import-partial-extension-blacklist` -> `at-import-partial-extension-disallowed-list`

Resolves #571


kristerkari:
Thanks @01taylop !

The changes look otherwise good, but the problem is that renaming a rule is a breaking change. 

Instead of just renaming a rule, the correct way would be to first deprecate the rule and add the same rule with the new name to a minor release to give people time to migrate to the new name. Then after some time the old rules could be removed in a major release.

kristerkari:
There's actually a quite "recent" example of deprecating a rule in this repo: #867

01taylop:
Thank you for the feedback @kristerkari ! I will checkout that PR and make the necessary changes to re-introduce the existing rules with a deprecation notice.

At least it was good to review the changes required to rename the rules :)

README.md
```diff
@@ -196,6 +196,8 @@ Please also see the [example configs](./docs/examples/) for special cases.
 These rules are deprecated — we won't fix bugs nor add options, and we will remove them in the next major release. We recommend you use a pretty printer (like Prettier) alongside Stylelint rather than these rules.
 
 - [`at-import-no-partial-leading-underscore`](https://github.com/stylelint-scss/stylelint-scss/blob/v5.2.1/src/rules/at-import-no-partial-leading-underscore/README.md): Disallow leading underscore in partial names in `@import`.
+- [`at-import-partial-extension-blacklist`](https://github.com/stylelint-scss/stylelint-scss/blob/v6.2.0/src/rules/at-import-partial-extension-blacklist/README.md): Specify a blacklist of disallowed file extensions for partial names in `@import` commands.
```
01taylop, line 199: I am assuming that 6.2.0 is the correct version to use here, as the current version is 6.1.0 and this PR seems like a minor change so would be tagged as 6.2.0?

01taylop:
@kristerkari Just got back from a long-haul flight, but I think this iteration is what you hoping for?

kristerkari:
Thanks! I'll try to get this re-review this weekend 👍

kristerkari:
Sorry for the delay, but I'll review this tonight 👍

kristerkari (approved):
Looks good 👍  Thanks!

PR #962, opened 2024-02-29
Refactor variable search in `no-unused-private-members`.

pamelalozano16:
Refactor variable search to match regex in atRules params and declaration values. This improves findings reliability by eliminating the need to list all edge cases.


kristerkari (approved):
Looks like a good fix, thanks!

PR #985, opened 2024-04-02
Add at-mixin-no-risky-nesting-selector rule.

### LINKED ISSUE #892: Add a lint for dangerous parent selector use in mixins
It's relatively common for authors to write selectors like `.context &` as a way of logically scoping the current style rule to a given context. However, this isn't very reliable in mixins, because the mixin may be included within another nested style rule, at which point the `.context` may be misplaced.

It would be useful to have a lint for this case to help authors ensure they're not accidentally making assumptions in their mixins. It should specifically warn for `&` in style rules when:

* They're within a `@mixin`.
* They're nested within another style rule.
* They _aren't_ at the beginning of a given complex selector. (That is, `& .foo` is fine, as is `.foo, & .bar`.)
* ~~They _aren't_ within `@at-root`, since that can't be affected by parent style rules anyway.~~ (This isn't accurate, `&` is still affected by outer style rules.)

It would even be possible to suggest a fix for this by pulling the parent selector into the child and making the child style rule a sibling. For example:

```scss
@mixin foo {
  .parent {
    color: blue;

    .context & {
      color: red;
    }
  }
}
```

could be adjusted to:

```scss
@mixin foo {
  .parent {
    color: blue;
  }

  .context .parent {
    color: red;
  }
}
```

pamelalozano16:
Fix #892.

### DIFF
--- a/src/rules/at-mixin-no-risky-nesting-selector/README.md
+++ b/src/rules/at-mixin-no-risky-nesting-selector/README.md
@@ -1,0 +1,144 @@
+# at-mixin-no-risky-nesting-selector
+
+Disallow risky nesting selectors within a mixin.
+
+If a mixin contains a parent selector within another style rule, and is used in a nested context, 
+the output selector may include the outermost parent selector in an unexpected way.
+
+In this example:
+```scss
+@mixin foo {
+  .a {
+    color: blue;
+    .b & {
+      color: red;
+    }
+  }
+}
+
+.c {
+  @include foo;
+}
+```
+
+The user may expect `.c` to go outside all selectors in `foo`:
+`.c .b .a {...}`
+
+But this outputs:
+```scss
+.c .a {
+  color: blue;
+}
+.b .c .a {
+  color: red;
+}
+```
+
+However, if we pull the parent selector into the child and make the child style rule a sibling:
+```scss
+@mixin foo {
+  .a {
+    color: blue;
+  }
+  .b .a {
+    color: red;
+  }
+}
+
+.c {
+  @include foo;
+}
+```
+
+It outputs:
+```scss
+.c .a {
+  color: blue;
+}
+.c .b .a {
+  color: red;
+}
+```
+
+This only occurs when a parent selector meets all of the following conditions:
+- Is within a `@mixin` rule.
+- Is nested within another style rule.
+- Is not positioned at the beginning of a complex selector.
+
+## Options
+
+### `true`
+
+The following patterns are considered warnings:
+
+```scss
+@mixin foo {
+  .bar {
+    color: blue;
+    .baz & {
+      color: red;
+    }
+  }
+}
+```
+
+```scss
+@mixin foo {
+  .bar {
+    color: blue;
+    .qux, .baz & .quux{
+      color: red;
+    }
+  }
+}
+```
+
+The following patterns are _not_ considered warnings:
+
+```scss
+.foo {
+  .bar {
+    color: blue;
+    .baz & {
+      color: red;
+    }
+  }
+}
+```
+
+```scss
+@mixin foo {
+  .bar {
+    color: blue;
+    & .baz {
+      color: red;
+    }
+  }
+}
+```
+
+```scss
+.bar {
+  color: blue;
+  .baz & {
+    color: red;
+  }
+}
+```
+
+```scss
+.foo {
+  color: blue;
+  & .bar, .baz & .qux {
+    color: red;
+  }
+}
+```
+
+```scss
+@mixin foo {
+  .baz & {
+    color: red;
+  }
+}
+```
--- a/src/rules/at-mixin-no-risky-nesting-selector/__tests__/index.js
+++ b/src/rules/at-mixin-no-risky-nesting-selector/__tests__/index.js
@@ -1,0 +1,206 @@
+"use strict";
+
+const { ruleName } = require("..");
+
+testRule({
+  ruleName,
+  config: [true],
+  customSyntax: "postcss-scss",
+
+  accept: [
+    {
+      code: `
+      .parent {
+        color: blue;
+    
+        & .b {
+          color: red;
+        }
+      }
+    `,
+      description: "Nested parent selector"
+    },
+    {
+      code: `
+      .parent {
+        color: blue;
+    
+        & .b, &.context {
+          color: red;
+        }
+      }
+    `,
+      description: "Nested parent selector in complex selector"
+    },
+    {
+      code: `
+      .bar {
+        & .parent {
+          color: blue;
+      
+          .context {
+            color: red;
+          }
+        }
+      }
+    `,
+      description: "Parent selector nested in another style rule"
+    },
+    {
+      code: `
+      @mixin foo {
+        &.context {
+          color: red;
+        }
+      }
+    `,
+      description: "Parent selector in mixin"
+    },
+    {
+      code: `
+      .bar {
+        .parent {
+          color: blue;
+      
+          & .context {
+            color: red;
+          }
+        }
+      }
+    `,
+      description: "Parent selector nested in more than one style rule"
+    },
+    {
+      code: `
+      .parent {
+        color: blue;
+    
+        .context & {
+          color: red;
+        }
+      }
+    `,
+      description: "Selector ending in parent selector"
+    },
+    {
+      code: `
+      .parent {
+        color: blue;
+    
+        .context & .b {
+          color: red;
+        }
+      }
+    `,
+      description: "Parent selector in the middle of complex selector"
+    },
+    {
+      code: `
+      .parent {
+        color: blue;
+    
+        & .b, .context & {
+          color: red;
+        }
+      }
+    `,
+      description: "Complex selector, one ending in parent selector"
+    },
+    {
+      code: `
+      @mixin foo {
+        .parent {
+          color: blue;
+      
+          & .context {
+            color: red;
+          }
+        }
+      }
+    `,
+      description: "Parent selector in mixin, nested, at the beginning"
+    },
+    {
+      code: `
+      @mixin foo {
+        .bar & {
+          color: red;
+        }
+      }
+    `,
+      description: "Parent selector in mixin, at the end"
+    },
+    {
+      code: `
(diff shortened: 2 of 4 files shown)


src/rules/mixins-no-risky-parent-selectors/README.md
```diff
@@ -0,0 +1,133 @@
+# mixins-no-risky-parent-selectors
+
+
+Parent selectors are not reliable in the following contexts:
+
+- Within a @mixin rule.
+- Nested within another style rule.
+- When they are not positioned at the beginning of a complex selector.
+- Within a @at-root rule.
```
nex3, line 9: In the issue I filed, I said that it _shouldn't_ be considered an error within `@at-root`. However, I think I was wrong about that—`&` within `@at-root` is still affected by outer parent selectors, so I think there should be no special handling for `@at-root` in either direction.

src/rules/mixins-no-risky-parent-selectors/README.md
```diff
@@ -0,0 +1,133 @@
...
+    .baz & {      
+/*       ↑
+    * Disallow this */
+      color: red;
+    }
+  }
+}
+```
+
+A more deterministic approach would be:
```
nex3, line 26: Nit: the behavior is always deterministic, the issue is just that it might be unexpected.

src/rules/mixins-no-risky-parent-selectors/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Options
+
+### `true`
+
+The following patterns are considered warnings:
+
+```scss
+@mixin foo {
+  & .baz {
```
nex3, line 46: This shouldn't produce a warning, because the parent selector is at the beginning of the style rule and also because it's not nested within another style rule.

src/rules/mixins-no-risky-parent-selectors/README.md
```diff
@@ -0,0 +1,133 @@
...
+}
+```
+
+```scss
+.bar {
+  .bar {
+    color: blue;
+      & .baz {
+          color: red;
+      }
```
nex3, line 58: Nit: Extra indentation.

src/rules/mixins-no-risky-parent-selectors/README.md
```diff
@@ -0,0 +1,133 @@
...
+    color: red;
+  }
+}
+```
+
+```scss
+.bar {
+  .bar {
+    color: blue;
+      & .baz {
```
nex3, line 56: This shouldn't produce a warning because it's not in a mixin. Same with examples below.

src/rules/mixins-no-risky-parent-selectors/README.md
```diff
@@ -0,0 +1,133 @@
...
+    color: red;
+  }
+}
+```
+
+```scss
+@mixin foo {
+  color: blue;
+  @at-root .b {
+    & {
```
nex3, line 95: This shouldn't produce a warning for the same reasons as above, and _also_ because it's in an `@at-root`.

nex3 (changes requested):

src/rules/mixins-no-risky-parent-selectors/README.md
```diff
@@ -0,0 +1,133 @@
+# mixins-no-risky-parent-selectors
+
+
+Parent selectors are not reliable in the following contexts:
```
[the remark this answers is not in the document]
pamelalozano16: Thanks for clarifying! I applied all modifications, updated the description and added an example.

src/rules/mixin-no-risky-parent-selectors/__tests__/index.js
```diff
@@ -0,0 +1,159 @@
...
+        }
+      }
+    `,
+      description: "Parent selector nested in selector within a mixin",
+      message:
+        "Unexpected nested parent selector in @mixin rule. (scss/mixin-no-risky-parent-selectors)"
+    },
+    {
+      code: `
+      @mixin foo {
```
nex3, line 145: Also test this case but with `& .qux` instead of just `.qux`

src/rules/mixin-no-risky-parent-selectors/index.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+  return false;
+}
+
+function hasNestedParentSelector(selectors) {
+  return selectors
+    .split(",")
+    .some(
+      selector =>
+        selector.includes("&") && !selector.replace(" ", "").startsWith("&")
```
nex3, line 33: I don't think this is quite right. Specifically, this won't produce an error for `& .foo &`, but it should (another test case to add).

nex3 (changes requested):

src/rules/mixin-no-risky-parent-selectors/index.js
pamelalozano16: Ohh right.. thanks for pointing this out!

src/rules/mixin-no-risky-parent-selectors/__tests__/index.js
```diff
@@ -0,0 +1,186 @@
...
+      
+          & .context {
+            color: red;
+          }
+        }
+      }
+    `,
+      description: "Parent selector in mixin, nested, at the beginning",
+      message:
+        "Unexpected nested parent selector in @mixin rule. (scss/mixin-no-risky-parent-selectors)"
```
nex3, line 123: Shouldn't the tests in the `accept` list _not_ have messages, since they aren't expected to throw errors?

src/rules/mixin-no-risky-parent-selectors/index.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+    parent = parent.parent;
+  }
+  return false;
+}
+
+function hasNestedParentSelector(selectors) {
+  return selectors
+    .split(",")
+    .some(selector => selector.includes("&") && /(.)+&/.test(selector));
```
nex3, line 31: I think this incorrectly will flag `.foo, &` as a nested selector because of the space before `&`. (Also, the parentheses in `(.)+&` are unnecessary—since you aren't using capture groups, it behaves the same as `.+&`).

nex3 (changes requested):

src/rules/mixin-no-risky-parent-selectors/__tests__/index.js
pamelalozano16: Right, it shouldn't

src/rules/mixin-no-risky-parent-selectors/index.js
pamelalozano16: Right, I actually had `selector.replace(' ','')` to remove spaces, not sure why I removed that.

nex3 (approved):

kristerkari:
Thanks! I'll need to review this, but after a quick look I noticed that the rule name starts with `mixin-`. There is already a category of rules with the prefix `at-mixin-`, and this rule should be put there. 👍

pamelalozano16:
@kristerkari Any thoughts on this? :)

kristerkari:
@pamelalozano16 also, did you notice my comment about the rule name (`mixin-no-risky-parent-selectors` -> `at-mixin-no-risky-parent-selectors`)?

src/rules/at-mixin-no-risky-parent-selectors/README.md
```diff
@@ -0,0 +1,142 @@
+# mixin-no-risky-parent-selectors
```
kristerkari, line 1: The rulename needs to be updated here.

src/rules/at-mixin-no-risky-parent-selectors/__tests__/index.js
```diff
@@ -0,0 +1,194 @@
...
+      code: `
+      @mixin foo {
+        .bar {
+          color: blue;
+          .baz & {
+            color: red;
+          }
+        }
+      }
+    `,
```
kristerkari, line 158: Please add `column`, `endColumn`, `endLine`, and `line` properties to all of these `reject` objects. That way we can verify that the error message is shown in a correct position.

src/rules/at-mixin-no-risky-parent-selectors/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+If a mixin contains a parent selector within another style rule, and is used in a nested context,
```
kristerkari, line 3: If we really want to comply with Stylelint's guides, then this "single-line description" should be in the form of "Disallow ...". https://stylelint.io/developer-guide/rules/#write-the-readme

kristerkari (changes requested):
Still one comment about the rule name since I'm trying to keep the rule names consistent. 

It seems like the most of the rules that start with the `no-` prefix are ending in a plural form (`no-dollar-variables`, `no-duplicate-mixins`, `no-global-function-names`, etc.).

The rules that start with something else seem to be mostly ending in a singular word. In this case it would mean that `-selectors` could be changed to `selector`.

Another thing is related to the naming of `&`.  There is already a rule called `selector-no-redundant-nesting-selector`. I'm not sure what is the term that Sass uses, but it might be beneficial to keep it consistent unless using `nesting-selector` makes the name unclear.

This would mean that the name would be `at-mixing-no-risky-nesting-selector`. What do you think, would that name work?

src/rules/at-mixin-no-risky-parent-selectors/README.md
kristerkari: There are examples in stylelint's repo:
https://github.com/stylelint/stylelint/tree/main/lib/rules/at-rule-no-vendor-prefix

pamelalozano16:
Yes! Did you mean `at-mixin-no-risky-nesting-selector` ? If yes, I agree that's fine.

kristerkari:
> Yes! Did you mean `at-mixin-no-risky-nesting-selector` ? If yes, I agree that's fine.

Yeah! Sorry, I had a typo in the name.

kristerkari (approved):
Looks good now 👍

kristerkari:
A link to this rule missing in the repo README, but I can add that after merging.

PR #988, opened 2024-04-19
Support at-rules in `no-global-function-names`

### LINKED ISSUE #593: no-global-function-names doesn't report errors inside of @if statements.. it works fine otherwise
### What steps are needed to reproduce the bug?

Run Stylint on the following `scss` file:

```scss
@mixin is($breakpoint) {
  // No error reported!
  @if map-has-key($breakpoints, $breakpoint) { 

   // Getting an error as expected
    color: map-get($colors, red);   

    @include _media($from: $breakpoint, $until: map.get($next, $breakpoint)) {
      @content;
    }
  }
}
```

### What Stylelint configuration is needed to reproduce the bug?

```json
{
  "extends": ["stylelint-config-standard-scss", "stylelint-config-prettier-scss"],
  "rules": {
    "string-quotes": "single",
    "function-name-case": null,
    "scss/at-mixin-pattern": null,
    "scss/dollar-variable-pattern": null,
    "scss/at-function-pattern": null
  }
}

```

### How did you run Stylelint?

CLI with `"stylelint \"**/*.scss\"`

### Which version of Stylelint are you using?

`14.5.0`

### What did you expect to happen?

I was expecting an error, since I'm using `map-has-keys` instead of `map.has-keys`

### What actually happened?

No error is reported

### Does the bug relate to non-standard syntax?

_No response_

### Proposal to fix the bug

_No response_

FloEdelmann:
Fixes #593.

### DIFF
--- a/src/rules/no-global-function-names/index.js
+++ b/src/rules/no-global-function-names/index.js
@@ -2,6 +2,7 @@
 
 const valueParser = require("postcss-value-parser");
 const { utils } = require("stylelint");
+const getAtRuleParams = require("stylelint/lib/utils/getAtRuleParams.cjs");
 const namespace = require("../../utils/namespace");
 const ruleUrl = require("../../utils/ruleUrl");
 
@@ -173,7 +174,15 @@ function rule(value) {
     }
 
     root.walkDecls(decl => {
-      valueParser(decl.value).walk(node => {
+      checkValue(decl, decl.value);
+    });
+    root.walkAtRules(atRule => {
+      const params = getAtRuleParams(atRule);
+      checkValue(atRule, params);
+    });
+
+    function checkValue(parentNode, value) {
+      valueParser(value).walk(node => {
         const cleanValue = node.value.replace(interpolationPrefix, "");
 
         // Verify that we're only looking at functions.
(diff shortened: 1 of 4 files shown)


src/rules/no-global-function-names/index.js
```diff
@@ -2,6 +2,7 @@
 
 const valueParser = require("postcss-value-parser");
 const { utils } = require("stylelint");
+const getAtRuleParams = require("stylelint/lib/utils/getAtRuleParams.cjs");
```
kristerkari, line 5: We should try to avoid importing from `stylelint`'s utils since they can change in future stylelint versions and cause a breaking change.

Could we instead create a copy of `getAtRuleParams` in this library?

kristerkari:
Thanks! Looks really good, but there is just one small thing that needs to be changed in my opinion. 👍

src/rules/no-global-function-names/index.js
FloEdelmann: Sure, done in [`febc4c6`](https://github.com/stylelint-scss/stylelint-scss/pull/988/commits/febc4c61667bcd71a429de0e3d8813743c07fa15).

kristerkari (approved):
Looks good 👍

PR #1010, opened 2024-06-05
Fix false positives in `at-mixin-no-risky-nesting-selector`.

### LINKED ISSUE #1009: False positive in `at-mixin-no-risky-nesting-selector`.
In a list of complex selectors with pseudo classes, within a mixin, and in a nested context, it shouldn't trigger a warning since both parent selectors are at the beginning.
Ex.
```scss
@mixin theme {
  .a {
    &:focus,
    &:hover {
      background-color: #0000;
    }
  }
}
```

pamelalozano16:
Fix #1009.


kristerkari:
Thanks!

PR #1019, opened 2024-07-08
Fix `partial-no-import` message to include more accurate rule name

ybiquitous:
By this change, the `partial-no-import` rule message will be properly associated with the rule name. See the examples below:

Before:

```sh-session
$ echo '@import "file.scss"' | npx stylelint

<input css ZECpLj>
       ✖  The 'partial-no-import' rule won't work if linting in a code string without an actual file

✖ 1 problem (1 error, 0 warnings)
```

After:

```sh-session
$ echo '@import "file.scss"' | npx stylelint

<input css PE6AAM>
  1:2  ✖  This rule won't work if linting in a code string without an actual file  scss/partial-no-import

✖ 1 problem (1 error, 0 warnings)
```

In addition, this change updates `jest-preset-stylelint` to use `testRule()` instead of `postcss`.
Because `testRule()` is closer to the actual use of `stylelint` and easier to avoid any potential troubles.
See https://github.com/stylelint/jest-preset-stylelint/releases/tag/7.1.0

### DIFF
--- a/src/rules/partial-no-import/__tests__/index.js
+++ b/src/rules/partial-no-import/__tests__/index.js
@@ -1,319 +1,204 @@
 "use strict";
 
 const path = require("path");
-const postcss = require("postcss");
 const rule = require("..");
-const { messages } = rule;
+const { messages, ruleName } = rule;
 
-function logError(err) {
-  console.log(err.stack); // eslint-disable-line no-console
-}
+testRule({
+  ruleName,
+  config: [true],
 
-test("No file specified", done => {
-  expect.assertions(2);
+  accept: [
+    {
+      description: "Import a file from non-partial .scss",
+      code: "@import 'file.scss';",
+      codeFilename: path.join(__dirname, "test.scss")
+    },
+    {
+      description: "Ignores empty imports (Sass will throw an error instead)",
+      code: "@import '';",
+      codeFilename: path.join(__dirname, "_test.scss")
+    },
+    {
+      description: "Ignores empty imports (Sass will throw an error instead) 2",
+      code: '@import " ";',
+      codeFilename: path.join(__dirname, "_test.scss")
+    },
+    {
+      description: "Import a file from CSS",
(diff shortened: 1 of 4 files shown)


src/rules/partial-no-import/__tests__/index.js
```diff
@@ -1,319 +1,204 @@
...
+    {
+      description: "Import from a non-partial SCSS-file.",
+      code: '@import "bootstrap/variables";\n@import "font-awesome/variables";',
+      codeFilename: path.join(__dirname, "_der", "variables.scss")
+    }
+  ],
+
+  reject: [
+    {
+      description: "No file specified",
```
ybiquitous, line 67: [note] Each `description` in all the `accept`/`reject` cases is associated with the old description in `test()`. For example, see https://github.com/stylelint-scss/stylelint-scss/pull/1019/files#diff-52a0082c93d02f6ded3b526b7cc8d60de78eda7eabe80bc7f8bb9d85b57e3cc8L12

ybiquitous:
For your information, this `partial-no-import` test would fail with the `stylelint` HEAD revision due to the direct use of `postcss`. That's why I recommend using `testRule()`.

```diff
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "postcss-less": "^6.0.0",
     "postcss-scss": "^4.0.9",
     "prettier": "^3.3.2",
-    "stylelint": "^16.6.1"
+    "stylelint": "https://github.com/stylelint/stylelint.git"
   },
   "files": [
     "src/**/*.js",
```

```sh-session
$ npm run --ignore-scripts test -- partial-no-import
...
Test Suites: 1 failed, 1 total
Tests:       9 failed, 11 passed, 20 total
Snapshots:   0 total
Time:        46.995 s
Ran all test suites matching /partial-no-import/i.
```

kristerkari:
Nice, thanks! 👍

PR #998, opened 2024-05-23
Add `load-partial-extension` and deprecate `at-import-partial-extension`.

pamelalozano16:
Fix #959.

PR #1021, opened 2024-07-10
Fix false positive for mixins in `load-partial-extension`

FloEdelmann:
Follow-up to #998. Without the code change, these new tests are failing.


kristerkari:
Thanks!

Issue #1020, opened 2024-07-10
Invalid rules for nesting & selectors for sass 1.77.7


Mister-Hope:
See https://github.com/sass/dart-sass/commit/7203d6538981df6422c686198da2497909fa11b5 and https://sass-lang.com/documentation/breaking-changes/mixed-decls/

So in a lot of cases, people are expected to use `& { }` now.

kristerkari:
Thanks for the heads up!

Looks like this change will concer the `selector-no-redundant-nesting-selector` rule.

Bigdragon13th:
Might be off-topic but related to this breaking change, and I can't find it myself.
Is there any rule to prevent people from putting declarations after nested rules?

MorevM:
@Bigdragon13th 

You can force declarations to be before or after the nested rules [via `order` rule of `stylelint-order` plugin](https://github.com/hudochenkov/stylelint-order/blob/master/rules/order/README.md)

Bigdragon13th:
@MorevM Thank you, I'll take a look

Mister-Hope:
This issue is two years old now. I'm not entirely sure if I tweaked some settings to work around it, but I consider this resolved since I haven't seen any new reports or references since August 2024.

Closing it.

PR #1022, opened 2024-07-10
selector-no-redundant-nesting-selector: Add support for Dart Sass deprecation

kristerkari:
Related to #1020

### DIFF
--- a/src/rules/selector-no-redundant-nesting-selector/__tests__/index.js
+++ b/src/rules/selector-no-redundant-nesting-selector/__tests__/index.js
@@ -163,6 +163,43 @@ testRule({
       }
       `,
       description: ":where, followed by a pseudo-element selector, issue #1000"
+    },
+    {
+      code: `
+      .example {
+        color: red;
+
+        a {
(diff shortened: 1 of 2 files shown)


Mister-Hope:
LGTM, but see https://github.com/sass/dart-sass/issues/2276#issue-2399880250

As we can revert this change back once the new behavior is activated, so this change might better be opt-in with a new rule option.

kristerkari:
> LGTM, but see [sass/dart-sass#2276 (comment)](https://github.com/sass/dart-sass/issues/2276#issue-2399880250)

Thanks, I'll need to read through the comments to understand what's going on 😄

kristerkari:
> As we can revert this change back once the new behavior is activated, so this change might better be opt-in with a new rule option.

Yeah, I can also put it behind a rule option for opt-in.

Mister-Hope:
Any update with this as it has been a while?

kristerkari:
I actually I forgot to follow up on this, sorry @Mister-Hope 

I still need to read through the 47 comments 😁

PR #1023, opened 2024-07-11
Add `declaration-property-value-no-unknown` support for SCSS.

### LINKED ISSUE #1017: Add `declaration-property-value-no-unknown` support for SCSS.
Add [`declaration-property-value-no-unknown`](https://github.com/stylelint/stylelint/tree/main/lib/rules/declaration-property-value-no-unknown) support for SCSS features and ignore SassScript.
E.g.
```scss
// Functions
@function get-color() {
   @return #336699;
}
.element {
  color: get-color(); // Unexpected unknown value "get-color" (declaration-property-value-no-unknown)
}

// Variables shorthand values
$var1: 1px;
a {
  margin: 2px $var1; // Cannot parse property value "2px $var1" for property "margin" (declaration-property-value-no-unknown)
}

// Hidden declarations
$rounded-corners: false;
.button {
  border: 1px solid black;
  border-radius: if($rounded-corners, 5px, null); // Cannot parse property value "if($rounded-corners, 5px, null)" for property "border-radius" (declaration-property-value-no-unknown)
}

// Nested properties
.b {
  font: {
    size: red; // No errors
  }
  font-size: red; // Unexpected unknown value "red" for property "font-size" (declaration-property-value-no-unknown)
}
```

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcID0SAEA1AhgJwJaYCMAbGAZ1VIAsJsAXSzAOylQDdMiBXMgHUYBJ22AIzxUwgA4APANx8+mVMHm0AtjgDmuRmIBM01IJzC5jAL7zGKVADFOjMLVwRGpPgAEAZvcfPGqDRhaAFpIIhoACgBKJT5UePdsIM5sfwBiAGYMgDZsgE480wtGPgA6GBJVGEZaWP9UMJoxQJDG7Giiy2sAOTJaGCgy7U8IYIlMQLr49WwtHVRMTloIKfjUAghaZdUxAAZTNdRliT1pA9RinloVEZqxZRLr2lJcAC8YMSSoUyvLx9uQi93p8Bp0SlY0AAJXBQWD+WBgIg4TBOFxuATYCD2WBQUI0RgwbCkMSeDikGA-RilAhLZb+B7xDbYWDYMSSKQUCBEGHrJFgADW5yZLOC2EwUFwnGJqFwngi-Ex2IGeNShNIABpUABWaSaxicIhEKJgvjWACSNUJEi5KL8fH4EiSnlwUjEPBAxAgAvdlPKulWEtIEiRAE8xNpuQTgmlgA6nS6zCaIagALKYCTo-htaUROJrR24GZh1BpXa7ADsBE8nnVefi5MgzBwxbS2TA5a15agtfq9c4YDAZGlaV0AA5MOWACxanuHKBMQKsktQMAZLXT2drADuOEY2g0YjS1bAwgrfGNlmptHpdfWmAFGkVzDx4SX6gkpRa8uzmoLRYv5iWEBjAgOqIDOiQ3SYFUCAgDAUjQcGMClKQYCkKQoHgC4zoaLBDKoO68H9MwbgIKgADat7uqQtAhiQkatNhuAaMENFMPOzKsWhpF5gAupu7rYAavBkfh8TugiSJiqijBjJiEiErRwTsFwMDBIwoz2PyGlbiUZG0EJMB5sUZiYY2OE2DQ6i0LBABWpAuJhsAZnhebUbR9HaIC3Hum6IBIv0NHugJIA0XRFReXijA4axtDsTguKoehvkEf5KJ9O6fCmWYQA)

pamelalozano16:
Fix #1017.

Added support for:
- Functions
- Variables shorthand values
- Hidden declarations
- Nested properties


src/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -0,0 +1,287 @@
+// NOTICE: This file is generated by Rollup. To modify it,
+// please instead edit the ESM counterpart and rebuild with Rollup (npm run build).
```
kristerkari, line 2: This comment can be removed since it's not relevant to this repository.

src/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -0,0 +1,287 @@
...
+"use strict";
+
+const cssTree = require("css-tree");
+const isPlainObject = require("is-plain-object");
```
kristerkari, line 6: These new dependencies should be added to the `package.json` as `dependencies` to avoid things breaking in case Stylelint removes them from their dependencies.

src/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -0,0 +1,287 @@
...
+      ...(secondaryOptions?.propertiesSyntax ?? {})
+    };
+    const typesSyntax = secondaryOptions?.typesSyntax ?? {};
+
+    /** @type {Map<string, string>} */
+    const typedCustomPropertyNames = new Map();
+
+    // Unless we tracked return values of declared functions, they're all valid.
+    root.walkAtRules("function", atRule => {
+      unsupportedFunctions.push(extractFunctionName(atRule.params)[1]);
```
kristerkari, line 93: Hmm... To make it easier to keep the rule up-to-date with the original rule, it would be a bit cleaner to keep the unsupported function check from the stylelint rule, and instead push/map these to a separate array.

src/utils/atKeywords.js
```diff
@@ -0,0 +1,64 @@
+// NOTICE: This file is generated by Rollup. To modify it,
+// please instead edit the ESM counterpart and rebuild with Rollup (npm run build).
```
kristerkari, line 2: These comments can be removed.

src/utils/isStandardSyntaxValue.js
```diff
@@ -0,0 +1,56 @@
+// NOTICE: This file is generated by Rollup. To modify it,
+// please instead edit the ESM counterpart and rebuild with Rollup (npm run build).
```
kristerkari, line 2: Same, these comments can be removed.

src/utils/uniteSets.js
```diff
@@ -0,0 +1,15 @@
+// NOTICE: This file is generated by Rollup. To modify it,
+// please instead edit the ESM counterpart and rebuild with Rollup (npm run build).
```
kristerkari, line 2: Same here, can be removed.

src/utils/validateObjectWithArrayProps.js
```diff
@@ -0,0 +1,35 @@
+// NOTICE: This file is generated by Rollup. To modify it,
+// please instead edit the ESM counterpart and rebuild with Rollup (npm run build).
```
kristerkari, line 2: ...and here, can be removed.

src/rules/declaration-property-value-no-unknown/README.md
```diff
@@ -0,0 +1,68 @@
+# declaration-nested-properties-no-divided-groups
+
+Disallow nested properties of the same "namespace" be divided into multiple groups.
```
kristerkari, line 3: This seems to be the wrong README file.

kristerkari (changes requested):
Thanks! I left a few comments mostly related to copying the rule from the stylelint repo to this one.

I'm wondering if there are any additional unit tests that we could copy here since the rule is using quite many additional util functions?

src/rules/declaration-property-value-no-unknown/README.md
pamelalozano16: Huh, I do not know how that got there. 🤔

src/rules/declaration-property-value-no-unknown/index.js
pamelalozano16: Do you mean export the function directly from Stylelint? 🤔 
This is how the function looked originally
```
function containsUnsupportedFunction(cssTreeNode) {
	return Boolean(
		cssTree.find(
			cssTreeNode,
			(node) => node.type === 'Function' && ['clamp', 'min', 'max', 'env'].includes(node.name),
		),
	);
}
```
I extracted that array into a global array and I push the function names to that same array.

src/rules/declaration-property-value-no-unknown/__tests__/index.js
```diff
@@ -0,0 +1,326 @@
...
+    {
+      code: "a { font-size: clamp(1rem, 1vw + 1rem, 3rem); }"
+    },
+    {
+      code: "a { font-size: min(1rem, 3rem); }"
+    },
+    {
+      code: "a { font-size: max(1rem, 3rem); }"
+    },
+    {
```
pamelalozano16, line 58: @kristerkari I added some SCSS support tests, let me know if these look good pls. :)

src/rules/declaration-property-value-no-unknown/__tests__/index.js
```diff
@@ -0,0 +1,326 @@
...
+      code: "a { filter: alpha(opacity=30); }",
+      message: messages.rejectedParseError("filter", "alpha(opacity=30)"),
+      line: 1,
+      column: 13,
+      endLine: 1,
+      endColumn: 30
+    },
+    {
```
pamelalozano16, line 186: @kristerkari this one too.

src/rules/declaration-property-value-no-unknown/index.js
kristerkari: Sorry for the confusing explanation, let me try again:

Since we are copying the rule from Stylelint's repo, it would be good to keep the original rule as untouched as possible and then make the additional changes on top of that. That way it would be easier to maintain the rule in case the original rule changes.

So what I'm suggesting is that the `containsUnsupportedFunction` would remain the same as in the original rule, you would push the function names to an array, and add another early return next to `if (containsUnsupportedFunction(cssTreeValueNode)) return;`

src/rules/declaration-property-value-no-unknown/__tests__/index.js
kristerkari: Looks good 👍

package.json
```diff
@@ -17,7 +17,9 @@
     "postcss-media-query-parser": "^0.2.3",
     "postcss-resolve-nested-selector": "^0.1.1",
     "postcss-selector-parser": "^6.1.0",
-    "postcss-value-parser": "^4.2.0"
+    "postcss-value-parser": "^4.2.0",
+    "css-tree": "2.3.1",
+    "is-plain-object": "5.0.0"
```
kristerkari, line 22: Can you please move these new dependencies to be before `known-css-properties` and run `npm install` to update `package-lock.json`. NPM will anyway re-order the dependencies when something is removed or added to dependencies.

src/rules/declaration-property-value-no-unknown/index.js
pamelalozano16: Ohh okay gotcha, makes sense. Thanks for explaining further :)

kristerkari (approved):
Thanks! Looks good now 👍

kristerkari:
Reminder for myself: add this rule to the repo README.

src/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -0,0 +1,297 @@
...
+          matchesStringOrRegExp(name, namePattern)
+        ) || [];
+
+      return valuePattern && matchesStringOrRegExp(value, valuePattern);
+    };
+
+    const propertiesSyntax = {
+      // Take a shallow clone as this object will be appended to.
+      ...(secondaryOptions?.propertiesSyntax ?? {})
+    };
```
Mouvedia, line 83: This rule is not being synchronised with [`declaration-property-value-no-unknown`](https://github.com/stylelint/stylelint/tree/main/lib/rules/declaration-property-value-no-unknown).
i.e. some false positives/negatives were fixed on our side

pamelalozano16: @kristerkari Is there a way we can extend the functionality directly from stylelint's [declaration-property-value-no-unknown](https://github.com/stylelint/stylelint/tree/main/lib/rules/declaration-property-value-no-unknown) instead of mirroring the rule?

Mouvedia: I am not sure if it will fit in your case but there's an utility called [`checkAgainstRule`](https://stylelint.io/developer-guide/plugins/#stylelintutilscheckagainstrule) that you could try. It has some [caveats](https://github.com/stylelint/stylelint/issues/7903) though.

BTW [this validator](https://github.com/stylelint-scss/stylelint-scss/pull/1023/files#diff-cdae24c897b01cf7307a0924e81240af5a8304185c36ad784e3c610077132040R54) is missing an argument.

kristerkari: Thanks @Mouvedia for pointing out the fixes done to the stylelint rule 👍 

> Is there a way we can extend the functionality directly from stylelint's [declaration-property-value-no-unknown](https://github.com/stylelint/stylelint/tree/main/lib/rules/declaration-property-value-no-unknown) instead of mirroring the rule?

Like @Mouvedia already mentioned, it's possible with `checkAgainstRule`, but using it is not an ideal solution. 

We already tried using `checkAgainstRule` in the `scss/function-no-unknown` rule. 

What ended up happening was that `stylelint`'s original `function-no-unknown` rule was updated with some fixes to ignore SCSS code, which then broke the `scss/function-no-unknown` rule when users updated to the new `stylelint` version.

Copy/pasting the code from `stylelint` is not the greatest solution either, but at least that way a change in a `stylelint` rule can not break a rule in `stylelint-scss`.

PR #1024, opened 2024-07-15
Add property-no-unknown support for nested declarations

### LINKED ISSUE #672: Nested property shorthand error
With SCSS we can write a shorthand property and write the rest of that property to the value of that property. That's fine, but *Stylelint* will throw an error if we assign a value to shorthand.

## Input

```scss
button {
    border: none {
/*  ^? Unexpected nested property "border"
    (scss/declaration-nested-properties)
    Stylelintscss/declaration-nested-properties */
        block-end: 2px solid red;
    /*  ^? Unexpected unknown property "block-end"
        (property-no-unknown)
        Stylelintproperty-no-unknown */
    }
}
```

## Output

```css
button {
  border: none;
  border-block-end: 2px solid red;
}
```

pamelalozano16:
Fix #672.

### DIFF
--- a/src/rules/property-no-unknown/__tests__/index.js
+++ b/src/rules/property-no-unknown/__tests__/index.js
@@ -9,6 +9,24 @@ testRule({
 
   accept: [
     {
+      code: `
+      button {
+        border: none {
+            block-end: 2px solid red;
+        }
+      }
+      `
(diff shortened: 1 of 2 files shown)


kristerkari:
Thanks!

PR #1041, opened 2024-08-19
Add load-no-redundant-path rule

### LINKED ISSUE #960: `load-no-redundant-path`
Having a relative path such as `@use './foo.scss';` becomes redundant because specifying `./` is unnecessary. This would be a similar rule to [load-no-partial-leading-underscore](https://github.com/stylelint-scss/stylelint-scss/tree/master/src/rules/load-no-partial-leading-underscore).

pamelalozano16:
Disallow paths starting with './' in `@import`, `@use`, `@forward`.
Fix #960.


kristerkari:
Thanks @pamelalozano16 !

I was thinking a bit about this rule, and it seems like we're getting to the point where the `load-` rules are becoming quite specific for allowing or disallowing a thing.

Would it actually make sense to add `load-path-allowed-list`/`load-path-disallowed-list` rules instead of adding this one? 

That way create a config that would disallow paths beginning with `./` using a regex, and other developers could use the same rule for other purposes.

Let me know what you think about that idea.

pamelalozano16:
@kristerkari Yes, having a `load-path-disallowed-list` seems like a good idea. It would indeed provide more flexibility and reduce the need specific rules. Although, I do have a couple of follow-up questions:

Would the introduction of `load-path-disallowed-list` mean that we would remove the existing `load-partial-extension` and `load-no-partial-leading-underscore` rules?
If so, to encourage best practices (disallowing partial extensions, leading underscores and redundant paths), would it make sense to have these regex disallowed by default when enabling `load-path-disallowed-list`?

Something like:
```js
config: [true, { disallowList: [/foo$/, "bar"], except: ["/./", /\".\//] }]
```
Basically, `disallowList` would extend the default list, and `except` is the option to disable partial extensions, leading underscores or redundant paths.

Looking forward to your thoughts on this!

kristerkari:
> Would the introduction of `load-path-disallowed-list` mean that we would remove the existing `load-partial-extension` and `load-no-partial-leading-underscore` rules?

My opinion is that it would be best to not do anything for those rules for now. We can definitely think about deprecating those rules in the near future, but in that case the replacements for those rules should be well documented.

> If so, to encourage best practices (disallowing partial extensions, leading underscores and redundant paths), would it make sense to have these regex disallowed by default when enabling `load-path-disallowed-list`?

I don't think that such a generic rule should have any defaults. It's best that the user has to always define some list to allow/disallow. Same seems to apply to the rules in the stylelint repo, e.g. https://github.com/stylelint/stylelint/tree/main/lib/rules/at-rule-disallowed-list

In my opinion the correct way to encourage best practices is to create a stylelint config with your preferred rules and options.

pamelalozano16:
Got it, that makes sense. Although, I’m thinking automatic fixes for this might be tricky unless we specify that the automatic fix *removes* the violation from the path (or replaces the provided regex with an empty string). 
I guess an advantage of having specific rules or default settings can make automatic fixes more straightforward. Also, it might not be obvious what the regex for redundant paths should be, which could make it harder to use.
What are your thoughts of a boolean option like `disallowRedundantPath` instead?

kristerkari:
> automatic fixes for this might be tricky unless we specify that the --fix option for this rule _removes_ the violation from the path (or replaces the provided regex with an empty string).

True, I haven't thought about that. I'm not sure if it's a good idea, but we could still keep having specific rules (like `load-no-partial-leading-underscore`)  that would only include a disallow list + autofix.

pamelalozano16:
Sounds good! So should I go ahead and implement `load-no-redundant`?

kristerkari:
> So should I go ahead and implement `load-path-disallowed-list`

Yeah, I think that that would make sense. I hope that it does not get too complicated if we want to use the dissallow list inside other rules.

PR #1042, opened 2024-08-23
Fix false positives in `declaration-property-value-no-unknown` for dollar vars and operators.

pamelalozano16:
- Support first-class operators.
- Support SCSS variables as arguments in nested functions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,6 @@
     "stylelint": "^16.0.2"
   },
   "dependencies": {
-    "common-tags": "^1.8.2",
     "css-tree": "2.3.1",
     "is-plain-object": "5.0.0",
     "known-css-properties": "^0.34.0",
@@ -23,6 +22,7 @@
     "postcss-value-parser": "^4.2.0"
   },
   "devDependencies": {
+    "common-tags": "^1.8.2",
     "cross-env": "^7.0.3",
     "eslint": "^8.57.0",
     "github-contributors-list": "^1.2.5",
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "stylelint": "^16.0.2"
   },
   "dependencies": {
+    "common-tags": "^1.8.2",
     "css-tree": "2.3.1",
     "is-plain-object": "5.0.0",
     "known-css-properties": "^0.34.0",
--- a/src/rules/declaration-property-value-no-unknown/index.js
+++ b/src/rules/declaration-property-value-no-unknown/index.js
@@ -38,7 +38,7 @@ const meta = {
   url: ruleUrl(ruleName)
 };
 
-const SYNTAX_PROPERTY = /^syntax$/i;
+const SYNTAX_DESCRIPTOR = /^syntax$/i;
 
 function extractFunctionName(inputString) {
   const matches = [...inputString.matchAll(/(?:\s*([\w\-$]+)\s*)?\(/g)].flat();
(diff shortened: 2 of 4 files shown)


src/utils/extractFunctionParams.js
```diff
@@ -0,0 +1,7 @@
+"use strict";
+
+module.exports = function (str) {
+  // Regex to match function calls and their parameters
+  const functionCallRegex = /(\w+)\s*\(([^()]*)\)/g;
+  return str.matchAll(functionCallRegex);
```
kristerkari, line 6: Not sure if this utility is doing the exact same thing, but could we use the existing `parseFunctionArguments` utility?

https://github.com/stylelint-scss/stylelint-scss/blob/master/src/utils/__tests__/parseFunctionArguments.js#L320

src/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -175,6 +178,18 @@ function rule(primary, secondaryOptions) {
...
         if (isIfStatement(value)) return;
+        if (isFunctionCall(value)) {
+          for (const i of extractFunctionParams(value)) {
+            if (i.some(val => isDollarVar(val))) return;
+          }
+        }
+        const operators = findOperators({ string: value });
+        if (operators) {
+          for (const i of operators) {
+            if (value.match(new RegExp("\\s*\\" + i.symbol + "\\s*", "g")))
```
kristerkari, line 189: How about:

```js
const unsupportedFunctions = ["clamp", "min", "max", "env"];
const mathOperators = ["+", "/", "-", "*", "%"];
....
```

```js
const operators = findOperators({ string: value }).map(o => o.symbol);

for (const operator of operators) {
  if (mathOperators.includes(operator)) {
    return;
  }
}
```

src/utils/extractFunctionParams.js
pamelalozano16: Right! Thanks

kristerkari:
Thanks! Looks really good now. 👍  

I still need to have another look at this tomorrow to review the fixes that you did in be163f4.

package.json
```diff
@@ -13,6 +13,7 @@
     "stylelint": "^16.0.2"
   },
   "dependencies": {
+    "common-tags": "^1.8.2",
```
kristerkari, line 16: Sorry that I missed this in the previous review round, but this should be in `devDependencies`, not in `dependencies`.

kristerkari (changes requested):

pamelalozano16:
I noticed declaration-property-value-no-unknown overlaps with unit-no-unknown and color-no-invalid-hex. E.g. [Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYAdAdugLgWyQCcBzAS03hQEZCY88BuLXSAGwkMoGIAzHrAL5YQAGhA9SrGADkkeOIhgAPOQAcpAOjABnbaPARME4ghAZMKFOhDKcMTFG3XKAbSyXL17TgCeU1uQ4ALSQRqTEQd5IDkRQ1u4oALoiCdaEAK5STghoCZ4gsGCsREg4pIZBqoQQqjCEvkEAbkis6TBBmBBB6ZgA1p0A7tg5OBkwKRYe1j2kwZ3dfYPDlKNtEx5WBuyEHV3kzQFQQQAWys4oqzAJQpgC+qHGAGIcBDimAFbahvqwqnqI5ny3j8MACmBw52sxTs3ms602wP8gRChmMkRw0SgsUhIGhMFhIEEIAEQA)

What are our thoughts on this?

kristerkari:
> What are our thoughts on this?

It seems to be something that is expected, and mentioned in `declaration-property-value-no-unknown` readme:

---

Screenshot 2024-09-11 at 22 42 27

---

https://github.com/stylelint/stylelint/tree/main/lib/rules/declaration-property-value-no-unknown

kristerkari:
@pamelalozano16 one last thing: could you also run `npm install` to update the `package-lock.json` after moving `common-tags` to dev dependencies?

kristerkari (approved):
Looks really good now, thanks! 👍

PR #1052, opened 2024-09-18
Add `color.channel` function

FloEdelmann:
It was added in Sass 1.78 (see https://sass-lang.com/documentation/modules/color/#channel). Using it without this PR causes [`function-no-unknown`](https://github.com/stylelint-scss/stylelint-scss/tree/master/src/rules/function-no-unknown) to report it as unknown.


kristerkari:
Thanks!

PR #1061, opened 2024-10-09
Fix false positives for imported custom functions in `scss/declaration-property-value-no-unknown`

pamelalozano16:
Currently imported custom functions are being flagged as `Unknown value`.
E.g.
```scss
// foo.scss
@function one() {
  @return 1;
}
```

```scss
// bar.scss
@import "foo";
.b {
  margin: one(); // Unexpected unknown value "one(" for property "margin" (scss/declaration-property-value-no-unknown).
}
```
This change checks whether the function exists in the CSS syntax (see: https://github.com/mdn/data/blob/main/css/syntaxes.json). If it doesn't exist, the function is tagged as a custom function, therefore skipped.

### DIFF
--- a/src/rules/declaration-property-value-no-unknown/index.js
+++ b/src/rules/declaration-property-value-no-unknown/index.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const cssTree = require("css-tree");
+const syntaxes = require("mdn-data/css/syntaxes");
 const isPlainObject = require("is-plain-object");
 const typeGuards = require("../../utils/typeGuards.js");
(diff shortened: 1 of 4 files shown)


src/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -1,6 +1,7 @@
 "use strict";
 
 const cssTree = require("css-tree");
+const mdnData = require("mdn-data");
```
kristerkari, line 4: Could we isntead import from the subpath?

```js
const syntaxes = require("mdn-data/css/syntaxes");
```

kristerkari (approved):
Looks good now 👍

PR #1062, opened 2024-10-10
Include explicit namespace variables in isDollarVar()

pamelalozano16:
Update `isDollarVar()` to include explicit namespace variables in multi-line values.
E.g.
```scss
@use foo;
.b {
  transition: background-color 0.3s 
  foo.$bar; // isDollarVar('foo.$bar') == true
}
```


kristerkari:
Thanks! Could you please also add a unit test(s) for this change?

PR #1068, opened 2024-10-16
Fix false positives for nested properties in nested decls

pamelalozano16:
In this example:
```scss
$acx-white: "#fff";
.b {
    background: {
      clip: padding-box;
      color: $acx-white;
    }
}
```

Both `padding-box` for `background-clip`, and `$acx-white` for `background-color` are being flagged as unknown values.


kristerkari:
Nice, thanks!

PR #1070, opened 2024-10-21
Add missing native functions

jhae-de:
This pull request adds some missing native CSS functions. In addition, the list of built-in module functions of Sass has been supplemented.


kristerkari:
Very nice, thanks @jhae-de !

PR #1071, opened 2024-10-21
Support nested properties with shorthand values in declaration-proper…

pamelalozano16:
…ty-value-no-unknown

### DIFF
--- a/src/rules/declaration-property-value-no-unknown/__tests__/index.js
+++ b/src/rules/declaration-property-value-no-unknown/__tests__/index.js
@@ -299,7 +299,7 @@ testRule({
       description: "Multi-line value is a explicit namespace variable."
     },
     {
-    code: `
+      code: `
     $acx-white: "#fff";
     .b {
         background: {
@@ -308,10 +308,10 @@ testRule({
         }
     }
     `,
-    description: "Nested properties in nested declarations."
-  },
-  {
-    code: `
+      description: "Nested properties in nested declarations."
+    },
+    {
+      code: `
     @use "foo";
     @function color(){
       @return 'blue';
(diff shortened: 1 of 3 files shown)


src/utils/validateTypes.js
```diff
@@ -61,15 +61,15 @@ function isIfStatement(value) {
  * @param {unknown} value
  * @returns {value is variable}
  */
-const isDollarVar = value =>
-  (value.length > 0 && value[0] === "$") || value.includes(".$");
+const isDollarVar = value => value.match(/\$[A-Za-z0-9_-]+/g) !== null;
```
kristerkari, line 64: I have to say that I like the non-regex version, both for simplicity, and possible perf reasons.

I have a feeling if we want to switch to a regex, then we should also add a proper unit tests for this function.

The question is, how much do we want to be manually parsing the variables from the value?

 I would prefer keeping the simpler version, but just making sure that the length is at least two chars: 

```js
const isDollarVar = value =>
  (value.length > 1 && value[0] === "$") || value.includes(".$");
```

src/rules/declaration-property-value-no-unknown/__tests__/index.js
```diff
@@ -323,8 +323,28 @@ testRule({
...
+    }
+    `,
+      description: "Nested properties and shorthand values."
+    },
+    {
+      code: `
+      @use "qux";
+      .b {
+        margin: $foo;
+        margin: (-$foo);
```
kristerkari, line 343: Related to my other comment, is `margin: (-$foo);` something that we should support in `isDollarVar`, or would it be better if the rule took care of parsing this value?

The reason why I'm asking is that it's quite easy to end up having to add more cases that include some exotic syntax that is valid SCSS if we choose to start supporting them.

kristerkari (changes requested):

src/rules/declaration-property-value-no-unknown/__tests__/index.js
pamelalozano16: Makes sense! Since the function is named isDollarVar, it should return whether the argument is strictly a variable (not if it's containing one).

kristerkari (approved):
Looks good now 👍

PR #1072, opened 2024-10-22
Unify rule messages

jhae-de:
All Stylelint rules and almost all rules in this project use error messages without a trailing punctuation mark. This pull request unifies all of these messages and removes trailing punctuation marks.


kristerkari:
Thanks @jhae-de ! These are the small, but important things 👍

PR #1077, opened 2024-11-02
Add `no-import` rule

qucumbah:
`@import` declaration is deprecated and expected to be removed in SASS 3.x.
This rule prohibits the usage of `@import`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- Added: `no-import` rule.
+
 # 6.8.1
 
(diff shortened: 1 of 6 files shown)


kristerkari:
Thanks @qucumbah !

I was wondering if there is a need for this rule, or could the same be done with Stylelint's already existing [`at-rule-disallowed-list`](https://github.com/stylelint/stylelint/tree/main/lib/rules/at-rule-disallowed-list) rule?

e.g.

```js
at-rule-disallowed-list: ['import']
```

qucumbah:
Yes, this rule does fit our needs, just couldn't find it earlier for some reason 🤷
Thanks, closing this PR then!

PR #1085, opened 2024-11-18
Fix multiline value parsing in declaration-property-value-no-unknown

pamelalozano16:
Strip multiline values before processing in `declaration-property-value-no-unknown`


kristerkari:
Looks good, but needs rebasing now that I merged #1071

Issue #1080, opened 2024-11-11
no-global-function-names does not allow using CSS round function


dgerhardt:
When the CSS `round` function is used, the `scss/no-global-function-names` rule reports an error.

More information:
* https://developer.mozilla.org/de/docs/Web/CSS/round
* https://drafts.csswg.org/css-values/#funcdef-round

kristerkari:
Thanks!

Looks like the Sass' support for CSS' `round` is not there yet, so I'm not sure how well it works currently.

- https://github.com/sass/sass/issues/3980
- https://github.com/sass/sass/issues/3803

Not sure should what the correct fix is. To remove `round` from the function name mapping in `no-global-function-names` rule perhaps?

Any suggestions @nex3 ?

nex3:
Sass supports plain-CSS `round()` just fine ([playground example](https://sass-lang.com/playground/#eJwzNLQqys8vUajmUgACXd280tyk1CIrBUMDa65arqT8lEqoVEFiSkpmXrqVQlF+aV6KRllikQZMtaYmUC0AA2UVrA==)). The only thing we don't yet support is `round(5.5)`, which was only recently added to browsers. This should be allowed by the lint.

PR #1097, opened 2024-12-30
Allow CSS round() function

davidwarrington:
Enables `round()` function in `no-global-function-names` since it's widely available https://developer.mozilla.org/en-US/docs/Web/CSS/round

Addresses #1080

Wasn't sure if any new tests were necessary for this PR. Please let me know if they are.


kristerkari:
Thanks for the fix!

PR #1103, opened 2025-01-07
Add `no-duplicate-load-rules` rule.

### LINKED ISSUE #1079: Add `no-duplicate-at-import-rules` to support other load rules.
`no-duplicate-load-rules` would disallow duplicate `@import`, `@use`, and `@forward` rules.

pamelalozano16:
Closes #1079

### DIFF
--- a/src/rules/no-duplicate-load-rules/README.md
+++ b/src/rules/no-duplicate-load-rules/README.md
@@ -1,0 +1,58 @@
+# no-duplicate-at-import-rules
+
+Disallow duplicate `@import` rules.
+
+<!-- prettier-ignore -->
+```css
+    @import "a.css";
+    @import "a.css";
+/** ↑
+ * These are duplicates */
+```
+
+The [`message` secondary option](../../../docs/user-guide/configure.md#message) can accept the arguments of this rule.
+
+## Options
+
+### `true`
+
+The following patterns are considered problems:
+
+<!-- prettier-ignore -->
+```css
+@import 'a.css';
+@import 'a.css';
+```
+
+<!-- prettier-ignore -->
+```css
+@import url("a.css");
+@import url("a.css");
+```
+
+<!-- prettier-ignore -->
+```css
+@import "a.css";
+@import 'a.css';
+```
+
+<!-- prettier-ignore -->
+```css
+@import "a.css";
+@import 'b.css';
+@import url(a.css);
+```
+
+The following patterns are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+@import "a.css";
+@import "b.css";
+```
+
+<!-- prettier-ignore -->
+```css
+@import url('a.css') projection;
+@import url('a.css') tv;
+```
--- a/src/rules/no-duplicate-load-rules/README.md
+++ b/src/rules/no-duplicate-load-rules/README.md
@@ -1,6 +1,6 @@
-# no-duplicate-at-import-rules
+# no-duplicate-load-rules
 
-Disallow duplicate `@import` rules.
+Disallow duplicate `@import`, `@use` and `@forward` rules.
 
 <!-- prettier-ignore -->
 ```css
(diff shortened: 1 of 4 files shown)


kristerkari:
Sorry for the delay. I have been taking a bit longer Christmas break from Open source, but I'll have a look at this now. :)

src/rules/no-duplicate-load-rules/README.md
```diff
@@ -0,0 +1,266 @@
...
+@use 'a' with (
+$black: #222,
+$border-radius: 0.1rem
+);
+@use 'b' with ($blue: #333);
+```
+
+## Optional secondary options
+
+### `includeForward: true`
```
kristerkari, line 112: What's the reason not to enable this by default?

src/rules/no-duplicate-load-rules/README.md
```diff
@@ -0,0 +1,266 @@
...
+
+<!-- prettier-ignore -->
+```scss
+@forward "a" show list-reset, $horizontal-list-gap;
+@forward "b" show $vertical-list-gap;
+```
+
+### `includeConfigurations: { "atRuleName": true }`
+
+Take into account configurations when checking for duplicate load rules.
```
kristerkari, line 186: Same here, what's the reason not to enable this functionality by default?

src/rules/no-duplicate-load-rules/index.js
```diff
@@ -0,0 +1,184 @@
...
+    const imports = {};
+    const hasExplicitNamespace = new RegExp(
+      /\s+(as|with|show|hide)\s+\(?[^;]*\)?/g
+    );
+    if (secondaryOptions?.includeForward) {
+      loadAtRules.push("forward");
+    }
+
+    root.walkAtRules(atRule => {
+      if (!loadAtRules.includes(atRule.name.toLowerCase())) {
```
kristerkari, line 53: `walkAtRules` accepts a regex as the first param, so we could possibly do this check there, but I'm not sure if it will be any cleaner (I guess it depends on if we'd want enable forward by default and something like `root.walkAtRules(/^(forward|import|use)$/i, atRule => {` )

kristerkari (commented):
A couple of small comments, other than that looks good 👍

src/rules/no-duplicate-load-rules/README.md
pamelalozano16: After discussing with the team offline, we’ve agreed that `@forward` should be required for the rule, but an error should only be triggered if there are two `@forward` rules with identical configurations, so I’ve removed the option entirely.

src/rules/no-duplicate-load-rules/README.md
pamelalozano16: We also concluded that since two `@import` and `@use` load rules with identical configurations already trigger a compilation error, this option isn’t needed. Therefore, I’ve removed it as well.

src/rules/no-duplicate-load-rules/index.js
pamelalozano16: Thanks! Done.

src/rules/no-duplicate-load-rules/index.js
```diff
@@ -0,0 +1,175 @@
...
+    const validOptions = utils.validateOptions(
+      result,
+      ruleName,
+      {
+        actual: primary
+      },
+      {
+        actual: secondaryOptions,
+        possible: {
+          includeForward: isBoolean,
```
kristerkari, line 31: You mentioned that you removed the secondary options, but they are still listed here.

src/rules/no-duplicate-load-rules/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+<!-- prettier-ignore -->
+```scss
+    @forward "a" with ($blue: #333);
+    @forward "a" with ($blue: #333);
+/** ↑
+ * With @forward, they are only considered duplicates if they have the exact same confuguration. */
+```
+
+The [`message` secondary option](../../../docs/user-guide/configure.md#message) can accept the arguments of this rule.
```
kristerkari, line 29: This should link to the `stylelint` repo, not this repo.

src/rules/no-duplicate-load-rules/README.md
```diff
@@ -0,0 +1,143 @@
+# no-duplicate-at-import-rules
```
kristerkari, line 1: The title has the wrong rule name.

kristerkari (changes requested):
A couple of further comments, after those: 👍

kristerkari (approved):

Issue #1090, opened 2024-11-26
Add `declaration-block-trailing-semicolon` support for nested properties and shorthand values.


pamelalozano16:
`declaration-block-trailing-semicolon` support for nested properties / shorthand values.

This example: 
```
div {
  border: solid #000 {
    width: 1px 1px 1px 7px;
  }
}
```
Triggers a false positive: `line 2, col 20 - Expected a trailing semicolon (declaration-block-trailing-semicolon)`

kristerkari:
I'm not sure how I feel about this one since Stylelint has deprecated all rules that have to do with styling. 

Before Prettier it was quite common to have stylistic linting rules, but since Prettier became popular, a lot of the stylistic stuff has been deprecated and moved into `stylelint-stylistic`. 

There has been discussion about it in this repo too (#814), but I have personally been to lazy to do the work to move the rules into a separate package.

Maybe it would be time to finally do that, and then you can open a PR against that package.

pamelalozano16:
Unfortunately, I don’t have the bandwidth to migrate all the rules to `stylelint-stylistic`, but I’m happy to move this PR once that package is available.

kristerkari:
Don't worry, I can take care of the migration when I have a free moment. I think that it would be beneficial to do the sepration in the same way that Stylelint did already a long time ago.

PR #1104, opened 2025-01-07
`declaration-block-trailing-semicolon` support for nested properties.

pamelalozano16:
Solves #1090


kristerkari:
Please check my comment in #1090. I think that it could be a good time to move stylistic rules in to a `stylelint-scss-stylistic` package and put this rule there.

PR #1113, opened 2025-01-24
Allow functions from explicit namespaces within other functions in `declaration-property-value-no-unknown`

### LINKED ISSUE #1112: False positive when calling a function from an explicit namespace within another function.
False positive when calling a function from an explicit namespace within another function from `declaration-property-value-no-unknown`.
E.g.
```scss
@use 'foo/bar';

.a {
  background-image: linear-gradient(
    to right,
    bar.get('blue') 27%,
    bar.get('red) 73%
  );
}
```
`Cannot parse property value "linear-gradient(
    to right,
    bar.get('blue') 27%,
    bar.get('red) 73%
  );" for property "background-image" (scss/declaration-property-value-no-unknown)`

pamelalozano16:
Fixes #1112.

### DIFF
--- a/src/rules/declaration-property-value-no-unknown/__tests__/index.js
+++ b/src/rules/declaration-property-value-no-unknown/__tests__/index.js
@@ -353,6 +353,20 @@ testRule({
         }
       `,
       description: "Multiline values with function calls."
+    },
+    {
+      code: `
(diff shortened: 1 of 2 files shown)


kristerkari:
Thanks!

PR #1123, opened 2025-02-19
Add support to Sass `rgba()` in `declaration-property-value-no-unknown`.

### LINKED ISSUE #1118: `declaration-property-value-no-unknown` breaking for a couple of valid cases
It seems like the rule `declaration-property-value-no-unknown` is breaking for a couple of cases that I can find with the error `✖  Cannot parse property value`

- `rgba` with a hex value
  
  ``` scss
  color: rgba(#888, 0.5);
  ```
- variables in multivalues properties
  
  ``` scss
  $foo: 1px;
  margin: 0 $foo;
  ```

pamelalozano16:
Fixes #1118.


kristerkari:
Thanks!

PR #1124, opened 2025-02-20
Check load duplicate rules by atRule name

### LINKED ISSUE #1121: [no-duplicate-load-rules] False positive when using @use and @forward for the same stylesheet
Given this stylesheet:
```scss
@use "foo";
@use "bar";

@forward "foo"; // <-- Unexpected duplicate load rule breakpoints (scss/no-duplicate-load-rules)

.test {
    margin: foo.$margin;
}
```

The `@forward` is reported as a duplicate when it shouldn't be.

From https://sass-lang.com/documentation/at-rules/forward/ (emphasis is mine):

> The rule is written `@forward` "<url>". It loads the module at the given URL just like `@use`, but it makes the [public](https://sass-lang.com/documentation/at-rules/use#private-members) members of the loaded module available to users of your module as though they were defined directly in your module. **Those members aren’t available in your module, though—if you want that, you’ll need to write a `@use` rule as well.** Don’t worry, it’ll only load the module once!

pamelalozano16:
Fixes #1121.

### DIFF
--- a/src/rules/no-duplicate-load-rules/__tests__/index.js
+++ b/src/rules/no-duplicate-load-rules/__tests__/index.js
@@ -346,6 +346,17 @@ testRule({
         $border-radius: 0.1rem
       );
       @forward 'src/list' with ($blue: #333);`
+    },
+    {
+      code: `
+      @use 'src/list';
+      @forward 'src/list' with ($blue: #333);`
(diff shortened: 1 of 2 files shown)


kristerkari:
Thanks!

PR #1130, opened 2025-02-22
Fix false negative for single quotes in scss/at-use-no-redundant-alias.

### LINKED ISSUE #1129: False negative for single quotes in `scss/at-use-no-redundant-alias`.
E.g. `@use 'sass:math' as math;` should raise `line 2, col 18 - Unexpected redundant namespace. (scss/at-use-no-redundant-alias)`, but it's ignored because of single quotes.

pamelalozano16:
Fixes #1129


kristerkari:
Looks good! 👍

PR #1142, opened 2025-04-01
Add `declaration-property-value-no-unknown` support for nested properties with shorthand values.

pamelalozano16:


kristerkari:
Thanks!

PR #1185, opened 2025-10-02
Handle mixin calls with content block arguments in `at-mixin-argumentless-call-parentheses` rule

### LINKED ISSUE #1184: Rule `at-mixin-argumentless-call-parentheses` fails for mixin calls with content block arguments
The `scss/at-mixin-argumentless-call-parentheses` rule does not correctly handle mixin calls with content block arguments and may fail report parentheses usage as expected. This leads to false negatives and inconsistent linting results.

**Stylelint configuration**
```javascript
{
  'scss/at-mixin-argumentless-call-parentheses': 'always',
}
```

**CSS code**
```css
@include my-mixin using($content-block-argument);
```

**Current behaviour**

The rule does not complain about the missing parentheses after the mixin name. Conversely, it also does not complain if the rule is configured with `never` and the mixin is called with parentheses.

**Expected behaviour**

The rule should report this as a problem:
```
Expected parentheses in mixin "my-mixin" call (scss/at-mixin-argumentless-call-parentheses)
```

The auto fix should result in:

```css
@include my-mixin() using($content-block-argument);
```

See [Sass documentation](https://sass-lang.com/documentation/at-rules/mixin/#passing-arguments-to-content-blocks).

jhae-de:
Fixes https://github.com/stylelint-scss/stylelint-scss/issues/1184

### DIFF
--- a/src/rules/at-mixin-argumentless-call-parentheses/__tests__/index.js
+++ b/src/rules/at-mixin-argumentless-call-parentheses/__tests__/index.js
@@ -11,72 +11,475 @@ testRule({
   accept: [
     {
       code: `
-      @include foo ;
-    `,
+        @include foo ;
+      `,
       description: "No parens; space after mixin name in a call."
     },
     {
       code: `
-      @include foo;
-    `,
+        @include foo;
+      `,
       description: "No parens; no space after mixin name in a call."
     },
     {
       code: `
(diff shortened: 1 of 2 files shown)


kristerkari (approved):
Amazing work with the unit tests 👍 

Also, sorry for the delay with this.

PR #1192, opened 2025-10-30
Add rule 'at-mixin-call-parentheses-space-before'

pioug:
Hello, I want to enforce a linter rule in a project that currently has a mix of

```scss
@include my-mixin();
@include my-mixin ();
```

I straight up copied `at-mixin-parentheses-space-before`. I have never maintained linter rules before, so I’m not sure whether that’s fine.

I was tempted to name it `at-include-parentheses-space-before`, but I saw there is `at-mixin-argumentless-call-parentheses`, so I named it `at-mixin-call-*` too.

### DIFF
--- a/src/rules/at-mixin-call-parentheses-space-before/README.md
+++ b/src/rules/at-mixin-call-parentheses-space-before/README.md
@@ -1,0 +1,58 @@
+# at-mixin-call-parentheses-space-before
+
+Require or disallow a space before `@include` parentheses.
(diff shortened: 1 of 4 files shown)


pioug:
@kristerkari No worries! I’ll just keep it as a local rule in my project for now. If you ever need help with maintenance, I’d be glad to contribute since we use this at BandLab and it’s working great for us.

PR #1195, opened 2025-11-12
Update `dollar-variable-no-missing-interpolation` to check custom properties, allow `--fix`

### LINKED ISSUE #597: Rule idea: Warn of using variables without interpolation in css custom properties
I've been bitten a few times by writing SCSS variables in CSS custom properties without interpolation. It would be great if we could lint and warn about this potential issue automatically.

Basically the same problem as here: https://github.com/stylelint-scss/stylelint-scss/issues/388

I read what `dollar-variable-no-missing-interpolation` is supposed to do and I agree with @kristerkari that it's not supposed to fix this issue. So I'm suggesting a new rule to handle this problem.

To re-use the example from that previous issue:

When I accidentally write:
```SCSS
.class {
  $var: 2rem;
  --gap: $var;

  display: grid;
  gap: var(--gap);
}
```
That gets rendered like this and does not work as I intended:
```CSS
.class {
  --gap: $var;

  display: grid;
  gap: var(--gap);
}
```
When I should have written:
```SCSS
.class {
  $var: 2rem;
  --gap: #{$var};

  display: grid;
  gap: var(--gap);
}
```
That gets rendered to:
```CSS
.class {
  --gap: 2rem;

  display: grid;
  gap: var(--gap);
}
```

If you have any other ideas how to tackle this issue with already existing rules, please let me know.

alexander-turner:
# Update `dollar-variable-no-missing-interpolation` to check custom properties, allow `--fix`

_Note: The original PR proposed adding a new rule. For future reference, I updated the description for the final design._

Enforces proper interpolation of SCSS variables in CSS custom properties (CSS variables). Without this interpolation, SCSS variables will be treated as literals and ignored by browsers.

Fixes #597.

## Problem

When SCSS variables are used directly in CSS custom properties without interpolation, they are output as literal strings instead of their values:

```scss
$spacing: 2rem;

// ❌ Wrong - outputs as "--gap: $spacing;" literally
.container {
  --gap: $spacing;
}

// ✅ Correct - outputs as "--gap: 2rem;"
.container {
  --gap: #{$spacing};
}
```

This is a common mistake that can be difficult to debug, as the CSS appears valid but doesn't work as intended.

## Solution

This PR updates `scss/dollar-variable-no-missing-interpolation` to:
- Detect SCSS variables used in custom properties without interpolation
- Provide automatic fixing via `--fix` (including for the rule's previous use cases)

### Invalid

```scss
a { --foo: $bar; }
a { --foo: calc(100% - $bar); }
a { --foo: calc($bar + $baz); }
```

### Valid

```scss
a { --foo: #{$bar}; }
a { --foo: calc(100% - #{$bar}); }
a { --foo: calc(#{$bar} + #{$baz}); }
a { --foo: var(--bar); }  // CSS variables are fine
a { color: $bar; }  // Regular properties are fine
```

## Testing

All tests pass.

To my knowledge, the rule follows all stylelint and stylelint-scss conventions:
- Uses standard naming pattern (`custom-property-no-missing-interpolation`)
- Implements `meta.fixable` for automatic fixing
- Uses `a { --foo: ... }` test conventions
- Comprehensive edge case coverage
- Proper use of `namespace()`, `ruleUrl()`, and stylelint utilities

## Configuration

```json
{
  "plugins": ["stylelint-scss"],
  "rules": {
    "scss/dollar-variable-no-missing-interpolation": true
  }
}
```

## Related

- Issue: #597
- Sass Documentation: https://sass-lang.com/documentation/style-rules/declarations/#custom-properties

### DIFF
--- a/src/rules/dollar-variable-no-missing-interpolation/README.md
+++ b/src/rules/dollar-variable-no-missing-interpolation/README.md
@@ -18,19 +18,19 @@ Sass variables that contain a custom identifier as a **string** require interpol
 For example, if you store a custom identifier as a string:
 
 ```scss
-$myVar: "myAnim";
+$foo: "bar";
 ```
 
 Then you must interpolate it when using it with custom identifier properties:
 
 ```scss
-animation: #{$myVar} 5s;
+animation: #{$foo} 5s;
 ```
 
 Without interpolation, Sass will compile the variable to a quoted string, producing invalid CSS:
 
 ```scss
-animation: "myAnim" 5s; // Invalid!
+animation: "bar" 5s; // Invalid!
 ```
 
 ### Examples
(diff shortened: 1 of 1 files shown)


kristerkari:
Thanks! I'll try to review as soon as possible 🙂

kristerkari:
@alexander-turner I quickly had a look at the changes, and they look really good!

However, I think that the Github issue that you linked (https://github.com/stylelint-scss/stylelint-scss/issues/597) is perhaps a bit misleading when it claims that my opinion is that this should added as a new rule. In the Github issue that it references, I simply closed the issue because the `dollar-variable-no-missing-interpolation` rule did not support CSS variables (the rule was created before CSS variables were a thing).

I don't really see a problem with expanding the `dollar-variable-no-missing-interpolation` rule to support CSS variables, unless there is some specific reason to add this as a new rule?

Let me know what you think, and let's discuss further if needed!

alexander-turner:
Thanks for the timely review! 

`dollar-variable-no-missing-interpolation` seems to exclusively address SCSS _string_ variables. IIUC, that rule only works when the SCSS variable to interpolate is a string --- often not true. So, while the name "`dollar-variable-no-missing-interpolation`" certainly has broad scope and while both rules certainly fix interpolations, their use cases are different (excepting the case where `--property: $var` for string `$var`). That pushed me towards making a new rule.

What are your thoughts?

alexander-turner:
@kristerkari Ping on this!

kristerkari:
Thanks @alexander-turner,

I agree that the existing `dollar-variable-no-missing-interpolation` rule and this one work differently, and I don't think that it would be in any way wrong to have them as separate rules.

Related to my previous comment, maybe I can explain my thinking a bit, and you can tell me if you agree or disagree with me.

I try to think about the rules more from a user's perspective rather than from the rule's perspective. 

The name of the rule `dollar-variable-no-missing-interpolation` suggests that there is an error/warning when a variable needs to be interpolated (you also pointed out that in your reply), so that's probably the reason why a user wants to enable the rule in their config. I would imagine that the users of the rule don't really care how the rule finds the variables that need interpolation.

So if we add this as a new rule, then I would think that the previously mentioned user would almost always want to enable both `custom-property-no-missing-interpolation` and `dollar-variable-no-missing-interpolation` rules (with the exception when a project does not use any CSS variables).

If instead we would add this as an additional functionality to the existing `dollar-variable-no-missing-interpolation` rule, then the already existing users would automatically get this functionality, and the new users would need to enable only one rule in their config.

I might be thinking about this in a totally wrong way, so let me know if it does not make sense. 😄

alexander-turner:
No worries! Sorry if I over-pinged, just excited to see this merge.

PR #1224, opened 2025-12-28
Change module type to ESM

jeddy3:
Replaces https://github.com/stylelint-scss/stylelint-scss/pull/931

The author hasn't been active on GitHub for a while, so it's easier to open a new PR that includes their commits. They'll be credited as a co-author.


kristerkari:
@jeddy3 Shouldn't the target branch be `v7`? I was still thinking of releasing one more patch release for v6, so it would be good to keep the ESM stuff out of master before that.

PR #1227, opened 2025-12-28
Prepare 7.0.0

### LINKED ISSUE #1215: Upcoming breaking changes in Stylelint 17.0.0
@kristerkari We wanted to give you a heads up about [some breaking changes coming in Stylelint 17.0.0](https://github.com/stylelint/stylelint/blob/8b509b7b5db529cf01595f0e9e089d981a68ee54/docs/migration-guide/to-17.md).

We've tried to keep them to a minimum, but some will affect `stylelint-scss`:

- removed the CommonJS Node.js API and Stylelint 17.0.0 will be an ES Module (ESM)
- change to standard spec-based nesting in a number of rules

This package will need to be an ESM to support the former (which was started in https://github.com/stylelint-scss/stylelint-scss/pull/931).

For the latter, we've [removed the `resolveNestedSelectors` option from `selector-class-pattern`](https://github.com/stylelint/stylelint/pull/8931) because it's incompatible with standard nesting, which doesn't concatenate strings. 

For example, this is valid in SCSS but invalid in CSS:

```css
.foo {
  &__bar {}
}
```

In the migration guide, we're encouraging the SCSS community to create their own plugin version of the rule. It would use nonstandard nesting under the hood to reinstate the option. It seems like this option is quite widely used SCSS authors. It can be done by repurposing [the removed](https://github.com/stylelint/stylelint/pull/8931) `flattenNestedSelectorsForRule.mjs‎` util that uses the `postcss-resolve-nested-selector` package to resolve selectors.

We've also changed the following rules to use standard nesting, which often desugars to `:is()` and therefore calculates specificity differently than if SCSS desugared them:

- `selector-max-specificity`
- `no-descending-specificity` 
- `no-duplicate-selectors`
- `selector-no-qualify-type`

For example:

```css
#foo
.bar {
  a {}
}
```

Is desugared in standard CSS to:

```css
:is(#foo, .bar) a {}
```

And the following in SCSS:

```css
#foo a {}
.bar a {}
```

We're recommending (in the migration guide and rule READMEs) that SCSS users turn off these rules in their configs, and you can do the same in the shared SCSS config, or create SCSS versions of these rules. However, the behaviour is only subtly different and not as big a change as removing the `resolveNestedSelectors` option from `selector-class-pattern`. If there's time to create only one plugin, we recommend it be a SCSS version of `selector-class-pattern`.

Feel free to drop any questions or queries here, and we'll do our best to answer them.

Thank you for supporting the SCSS community with the plugin!

jeddy3:
Closes #1215 

@kristerkari When you have a moment, it'd be great to double-check and approve this. It's `v7` set to merge to `master` when we're ready to release `7.0.0`.

> [!NOTE]
Merge. DO NOT SQUASH to maintain commits and PR history.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,14 +25,14 @@
         "github-contributors-list": "^1.2.5",
         "husky": "^9.1.7",
         "jest": "^30.2.0",
-        "jest-preset-stylelint": "^8.0.0",
+        "jest-preset-stylelint": "^9.0.0",
         "lint-staged": "^16.2.7",
         "np": "^10.2.0",
         "postcss": "^8.5.6",
         "postcss-less": "^6.0.0",
         "postcss-scss": "^4.0.9",
         "prettier": "^3.7.4",
-        "stylelint": "16.26.1"
+        "stylelint": "^17.0.0"
       },
       "engines": {
         "node": ">=20.19.0"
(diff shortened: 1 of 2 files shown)


kristerkari (approved):
Thanks a lot @jeddy3 for taking care of this! 👍 

I already reviewed the separate PRs, and I can't think of anything that would be missing.

I guess the only thing would be try to verify that after these changes this package works as ESM-only version, but I'm not sure if there is some really easy way to do that.

Other than that I think that this is good to go as soon as Stylelint 17 is out.

PR #1260, opened 2026-04-14
Fix CJS/ESM interop for postcss-media-query-parser import for Bun compatability

andyexeter:
When running under Bun, the default import of postcss-media-query-parser resolves directly to the parseMedia function rather than a module object with a .default property. This causes a `TypeError: mediaQueryParser is not a function` error at runtime when linting.

This PR adds a fallback to the initial import if the .default property is not present.


kristerkari:
Thanks! I'll just quickly need to try this out with Bun before merging.

PR #1263, opened 2026-04-28
Add support for `postcss-scss` in `declaration-nested-properties-no-divided-groups` rule

### LINKED ISSUE #1262: Add support for `postcss-scss` in `declaration-nested-properties-no-divided-groups` rule
I’ve always wondered why the example [documented](https://github.com/stylelint-scss/stylelint-scss/tree/v7.0.0/src/rules/declaration-nested-properties-no-divided-groups) in the `declaration-nested-properties-no-divided-groups` rule doesn’t trigger an error when using the `stylelint-config-standard-scss` configuration, even though the rule is enabled there (specifically in the `stylelint-config-recommended-scss` configuration):

```scss
a {
  background: url(img.png) center {
    size: auto;
  }
  background: {
    repeat: no-repeat;
  }
}
```

I finally had some time to investigate the cause. It turns out that the `stylelint-config-recommended-scss` configuration sets `postcss-scss` as custom syntax, but the rule doesn’t support that syntax at all. You can reproduce this by adding `customSyntax: "postcss-scss"` to the `reject` tests. In that case, all of these tests fail.

jhae-de:
Closes https://github.com/stylelint-scss/stylelint-scss/issues/1262

I also refactored the rule a bit to improve its performance and readability.

### DIFF
--- a/src/rules/declaration-nested-properties-no-divided-groups/__tests__/index.js
+++ b/src/rules/declaration-nested-properties-no-divided-groups/__tests__/index.js
@@ -2,33 +2,38 @@ import rule from "../index.js";
 
 const { ruleName, messages } = rule;
 
-testRule({
-  ruleName,
-  config: [true],
-  customSyntax: "postcss-scss",
-  skipBasicChecks: true,
+const syntaxVariants = [
+  { syntaxLabel: "default syntax" },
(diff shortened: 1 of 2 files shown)


kristerkari:
Thanks!

PR #1265, opened 2026-04-28
Add support for `postcss-scss` in `declaration-nested-properties` rule

### LINKED ISSUE #1264: Add support for `postcss-scss` in `declaration-nested-properties` rule
The `declaration-nested-properties` rule does not report an error if you extend the `stylelint-config-standard-scss` configuration and enable it with the "never" option.

The following [documented](https://github.com/stylelint-scss/stylelint-scss/tree/v7.0.0/src/rules/declaration-nested-properties#never) examples should report an error, but they don't:

```scss
a {
  margin: {
    left: 10px;
  }
}
```

```scss
a {
  background: red {
    repeat: no-repeat;
  }
}
```

Similar to the issue described [here](https://github.com/stylelint-scss/stylelint-scss/issues/1262), `postcss-scss` support is missing. There is also a corresponding commented-out `reject` [test case](https://github.com/stylelint-scss/stylelint-scss/blob/v7.0.0/src/rules/declaration-nested-properties/__tests__/index.js#L361) that fails when enabled.

jhae-de:
Closes https://github.com/stylelint-scss/stylelint-scss/issues/1264

I also refactored the rule a bit to improve its performance and readability.

### DIFF
--- a/src/rules/declaration-nested-properties/__tests__/index.js
+++ b/src/rules/declaration-nested-properties/__tests__/index.js
@@ -2,37 +2,42 @@ import rule from "../index.js";
 
 const { ruleName, messages } = rule;
 
+const syntaxVariants = [
+  { syntaxLabel: "default syntax" },
+  { syntaxLabel: "postcss-scss", customSyntax: "postcss-scss" }
+];
+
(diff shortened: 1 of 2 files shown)


kristerkari:
Very nice, thanks!

PR #1287, opened 2026-05-16
Fix missing interpolation autofix for Sass functions in custom properties

### LINKED ISSUE #1228: false positive when passing a string to a function
Here is a function helper to encode svg string.

```scss
@function encode-svg($svg) {
  $encoded: "";
  $slice: 2000;
  $index: 0;
  $loops: math.ceil(math.div(string.length($svg), $slice));

  @for $i from 1 through $loops {
    $chunk: string.slice($svg, $index, $index + $slice - 1);
    $chunk: str-replace($chunk, '"', "'");
    $chunk: str-replace($chunk, "%", "%25");
    $chunk: str-replace($chunk, "#", "%23");
    $chunk: str-replace($chunk, "{", "%7B");
    $chunk: str-replace($chunk, "}", "%7D");
    $chunk: str-replace($chunk, "<", "%3C");
    $chunk: str-replace($chunk, ">", "%3E");
    $encoded: #{$encoded}#{$chunk};
    $index: $index + $slice;
  }

  @return url("data:image/svg+xml;utf8,#{$encoded}");
}
```

Generally speaking, the `$svg` is expected to be filled with quote strings for sure.

Image

This is absolutely a false positive, as it makes no sense to use a function like this. The correct way should be

```scss
:root {
  --a: #{svg.encode-svg($icon)};
}
```

The plugin should suggest interpolation the whole function, not the param. Since it has an "auto fix", this could be pretty annoying, as a wrong fix is applied instead.

sjh9714:
## Summary

Fixes the autofix for Sass function calls in CSS custom properties.

## What changed

- Added regression tests for Sass function calls in custom property values.
- Updated the autofix to interpolate the full Sass function call when a variable appears inside one.
- Preserved native CSS function behavior such as `calc($size * 2)`.

## Why

Issue #1228 reports that `svg.encode-svg($icon)` is fixed as though only the argument needs interpolation, but the correct custom property value is `#{svg.encode-svg($icon)}`.

## Testing

- `npm run jest -- src/rules/dollar-variable-no-missing-interpolation/__tests__/index.js --runInBand`
- `npm run lint`
- `npm test`

Fixes #1228

### DIFF
--- a/src/rules/dollar-variable-no-missing-interpolation/__tests__/index.js
+++ b/src/rules/dollar-variable-no-missing-interpolation/__tests__/index.js
@@ -758,6 +758,73 @@ testRule({
     },
     {
       code: `
+      @use 'svg';
+
+      $external-link-icon: "<svg></svg>";
+
+      :root {
+        --external-link-icon: svg.encode-svg($external-link-icon);
+      }
+      `,
+      fixed: `
+      @use 'svg';
+
+      $external-link-icon: "<svg></svg>";
+
+      :root {
+        --external-link-icon: #{svg.encode-svg($external-link-icon)};
+      }
+      `,
+      message: messages.rejected("--external-link-icon", "$external-link-icon"),
+      description:
+        "should fix Sass function calls in CSS custom properties by interpolating the function"
+    },
+    {
+      code: `
+      @use 'sass:map';
+      @use 'sass:meta';
+
+      $theme: (font: serif);
+
+      :root {
+        --font: meta.inspect(map.get($theme, font));
+      }
+      `,
+      fixed: `
+      @use 'sass:map';
+      @use 'sass:meta';
+
+      $theme: (font: serif);
+
+      :root {
+        --font: #{meta.inspect(map.get($theme, font))};
+      }
+      `,
+      message: messages.rejected("--font", "$theme"),
+      description:
+        "should fix nested Sass function calls in CSS custom properties by interpolating the outer function"
+    },
+    {
+      code: `
+      $size: 10px;
+
+      :root {
+        --custom-size: calc($size * 2);
+      }
+      `,
+      fixed: `
+      $size: 10px;
+
+      :root {
+        --custom-size: calc(#{$size} * 2);
+      }
+      `,
+      message: messages.rejected("--custom-size", "$size"),
+      description:
+        "should fix variables inside native CSS functions in CSS custom properties"
+    },
+    {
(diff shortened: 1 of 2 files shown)


src/rules/dollar-variable-no-missing-interpolation/index.js
```diff
@@ -86,34 +87,87 @@ function isStringValue(value) {
...
+
+      if (node.nodes) {
+        walkNodes(node.nodes, nextFunctionStack);
+      }
+    }
+  }
+
+  walkNodes(valueParser(value).nodes);
+
+  return ranges
```
kristerkari: This sorting and filtering of the `ranges` here does not seem to affect the functionalty at all.

Could we just do `return ranges;` and remove the `.sort` and `.filter` calls?

src/rules/dollar-variable-no-missing-interpolation/index.js
```diff
@@ -86,34 +87,87 @@ function isStringValue(value) {
...
+    }
+  }
+
+  return functionStack
+    .slice(nativeFunctionBoundaryIndex + 1)
+    .find(functionNode => !isNativeCssFunction(functionNode.value));
+}
+
 /**
  * Wrap uninterpolated SCSS variables with interpolation syntax.
```
kristerkari: This comment is now out-of-date with the function that got changed. Could you please also update this comment block to match the new `getInterpolationRanges` function?

src/rules/dollar-variable-no-missing-interpolation/__tests__/index.js
```diff
@@ -756,6 +756,73 @@ testRule({
...
+
+      :root {
+        --custom-size: calc($size * 2);
+      }
+      `,
+      fixed: `
+      $size: 10px;
+
+      :root {
+        --custom-size: calc(#{$size} * 2);
```
kristerkari, line 857: Would you mind adding the `fixed` test cases to the `accepted` tests for the rule? That way we can make sure that they are not getting flagged when the code is already correct.

kristerkari (changes requested):
Looks good, just some small comments 👍

sjh9714:
Thanks for the review. I pushed a follow-up commit addressing the three comments: simplified `getInterpolationRanges`, updated the comment, and added the fixed cases to `accept` coverage.

kristerkari (approved):
Great work, thanks a lot! 👍

PR #1289, opened 2026-05-27
Add namespaced functions support to 'declaration-property-value-no-unknown'

### LINKED ISSUE #1288: False positive: `scss/declaration-property-value-no-unknown` throws parse error on Polymer mixins and namespaced functions
### What minimal example or steps are needed to reproduce the bug?

**Namespaced module function inside a CSS property:**
```scss
.element {
  background-image: linear-gradient(
    45deg,
    css-vars.plx-var(background, hover) 20%,
    transparent 20%
  );
}
```

### Which Stylelint-related dependencies are you using?
stylelint: 17.12.0

stylelint-scss: 6.11.1

### What did you expect to happen?
The linter should recognize this as non-standard or valid custom syntaxes (SCSS module namespaces) and either ignore it or pass validation.

### What actually happened?
The `css-tree` parser fails to parse this function, resulting in a rejectedParseError and causing the linter to fail with the following false positives:
```
Cannot parse property value "linear-gradient( 45deg, css-vars.plx-var(background, hover) 20%, transparent 20% )" for property "background-image" (scss/declaration-property-value-no-unknown)
```

### Do you have a proposal to fix the bug?
Yes. We can add  anearly return in the rule before cssTree.parse is called to intercept these specific non-standard formats before the parser chokes on it.

Adding the following helper functions and checks resolves the issue:

```JavaScript
// Add this check before `cssTree.parse(value, ...)`
if (containsNamespacedFunction(value)) return;

// Helper function to catch namespaced module functions `namespace.function()`
/**
 * Check if a value contains a namespaced function call.
 * @param {string} value
 * @returns {boolean}
 */
function containsNamespacedFunction(value) {
  return /[\w-]+\.[\w-]+\(/.test(value);
}
```

pamelalozano16:
Closes #1288.

### DIFF
--- a/src/rules/declaration-property-value-no-unknown/__tests__/index.js
+++ b/src/rules/declaration-property-value-no-unknown/__tests__/index.js
@@ -618,6 +618,34 @@ testRule({
         }
       `,
       description: "Nested properties and shorthand values."
+    },
+    {
+      code: `
+        .element {
+          background-image: linear-gradient(
+            45deg,
+            css-vars.var(background, hover) 20%,
+            transparent 20%
+          );
+        }
+      `,
+      description: "Namespace function call nested in linear-gradient."
+    },
+    {
+      code: `
+        .element {
+          background-color: css-vars.var(background, hover);
+        }
+      `,
+      description: "Namespace function call as declaration value."
+    },
+    {
+      code: `
+        border: solid #000 {
+          width: 1px 1px 1px 7px;
+        }
+      `,
+      description: "Nested properties and shorthand values."
     }
   ],
 
--- a/src/rules/declaration-property-value-no-unknown/index.js
+++ b/src/rules/declaration-property-value-no-unknown/index.js
@@ -313,6 +313,8 @@ const rule = (primary, secondaryOptions) => {
 
         if (hasDollarVarArg(value)) return;
 
+        if (containsNamespacedFunction(value)) return;
+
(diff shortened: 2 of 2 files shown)


kristerkari:
Thanks!

src/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -446,6 +448,16 @@ function containsCustomFunction(cssTreeNode) {
   );
 }
 
+// 3. Helper function to catch namespaced module functions `namespace.function()`
```
kristerkari: Minor thing, but I don't think that this extra comment starting with 3. is needed, or if it is, then it could be moved inside the JSDoc block.

src/rules/declaration-property-value-no-unknown/__tests__/index.js
```diff
@@ -611,6 +611,34 @@ testRule({
...
+      `,
+      description: "Namespace function call nested in linear-gradient."
+    },
+    {
+      code: `
+        .element {
+          background-color: css-vars.var(background, hover);
+        }
+      `,
+      description: "Namespace function call nested in linear-gradient."
```
kristerkari: Also here you could fix the comment since the function call is not nested and there is no linear-gradient.

kristerkari (changes requested):

kristerkari (approved):
Looks good now, thanks! 👍

PR #1328, opened 2026-08-04
Fix `operator-no-unspaced` false positive for CSS `type()`

### LINKED ISSUE #1310: Bug: `operator-no-unspaced`: false positive for CSS `type()`
The rule `operator-no-unspaced` returns false positive when using CSS [`type()`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/type):

```css
attr(data-color type(<color>))
/*                   ↑     ↑
Expected single space before ">"
Expected single space after ">"
*/
```

rkdfx:
Closes #1310

`<` and `>` were treated as comparison operators unconditionally, so the angle brackets in `attr(data-color type(<color>))` got flagged. Skip them inside a `type()` call.

Fixes the shared value parser, so `operator-no-newline-before/after` and `declaration-property-value-no-unknown` are covered too.


kristerkari:
Thanks for this! 

I already went and merged #1329, which was very similar to this, but without the looping. 

I appreciate your contribution. 👍  Just let me know if I missed something in the other PR compared to this one.

PR #1329, opened 2026-08-04
Fix false positive in operator-no-unspaced for CSS type() function arguments

### LINKED ISSUE #1310: Bug: `operator-no-unspaced`: false positive for CSS `type()`

l46983284-cpu:
# Fix false positive in `operator-no-unspaced` for CSS `type()` function arguments

Fixes #1310.

## Problem

`operator-no-unspaced` reported false positives for angle brackets inside the CSS
[`type()`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/type)
function, e.g.:

```css
a { color: attr(data-color type(<color>)); }
```

```
Expected single space before "<"
Expected single space after "<"
Expected single space before ">"
Expected single space after ">"
```

## Fix

In `findOperators` (`src/utils/sassValueParser/index.js`), `<` and `>` were treated as
comparison operators regardless of context. Added an `isInsideTypeFunctionCall` check:
a character is considered inside `type(...)` when the nearest unclosed `(` before it
belongs to a `type(` call. Angle brackets inside `type()` arguments denote a CSS data
type, not a comparison, so they are now skipped.

Real comparisons are still reported, including ones right after a `type()` call
(`type($var)< 1`), inside other functions (`prototype($var< 1)`), and in `@if`
conditions.

## Tests

- 8 new cases in `src/rules/operator-no-unspaced/__tests__/index.js`: the exact issue
  case, other data types (`<length>`), combined types (`<length> | <percentage>`),
  `attr()` fallback, spaces inside parens, uppercase `TYPE(<COLOR>)`, plus two reject
  cases proving comparisons adjacent to `type()` are still flagged.
- Rule suite: 346 passed, 2 skipped (was 338 passed, 2 skipped).
- Full suite (`jest --coverage`): 85 suites, 3284 passed, 14 skipped.


kristerkari:
Thanks for this! Looks good, but I will have look at the regex that you added to see what exactly it is targeting before merging this.

Also very nice that you added so many test cases!

kristerkari (approved):
Looks good 👍

PR #1333, opened 2026-08-08
Fix: report disallowed functions in at-rule params

### LINKED ISSUE #1240: Bug: `scss/function-disallow-list` ignores at-rules within which a function might be called
The `scss/function-disallow-list` misses function for the disallowed function inside at-rules that are not `@return`. For example, using the rule with the same `["random"]` configuration (rejecting calls to the `random` function) as in the tests, you'd expect the following pieces of code to be rejected (though it should probably apply to any at-rule, as functions may be used to create params for any CSS at-rule, and that list is ever growing).

## Inside an `@if`

```scss
@use "sass:math";

.a {
  @if math.random(100) > 50 {
    margin-left: 10px;
  } @else {
    margin-left: 20px;
  }
}
```

### Inside an `@media`

```scss
@use "sass:math";

@media (width >= #{math.random(100)}) {
  .a {
    margin-left: 10px;
  }
}
```

### Inside an `@include`

```scss
@use "sass:math";

@mixin min-width($value) {
  @media (width >= #{$value}) {
    @content;
  }
}

@include min-width(math.random(100)) {
  .a {
    margin-left: 10px;
  }
}
```

### Inside an `@mixin`

```scss
@use "sass:math";

@mixin min-width($value: math.random(100)) {
  @media (width >= #{$value}) {
    @content;
  }
}

@include min-width() {
  .a {
    margin-left: 10px;
  }
}
```

### Inside an `@function`

```scss
@use "sass:math";

@function min-width($value: math.random(100)) {
  @media (width >= #{$value}) {
    @content;
  }
}

@media #{min-width()} {
  .a {
    margin-left: 10px;
  }
}
```

rkdfx:
The rule only walked declaration values and `@return`, so disallowed functions in `@if`, `@else if`, `@media`, `@each`, `@include` arguments and `@mixin`/`@function` default values went unreported, now every at-rule's params are checked. Because the name in `@function foo()`, `@mixin foo()` and `@include foo()` parses as a function node, the first one is skipped for those three at-rules (before the native-CSS check) while still descending into its arguments, so mixin defaults and include arguments are still caught and declarations aren't falsely flagged.

Closes #1240.


kristerkari:
Sorry for taking so long, but looks good 👍

